PR #41, opened 2021-01-26
Fix sample and documentation code regarding auto registration

### LINKED ISSUE #40: Reflective registration of producers/consumers in samples seem off
I'm trying to use the `.Do(...)` reflective registration of producers and consumers as outlined in the samples repository of the WebApi: https://github.com/zarusz/SlimMessageBus/blob/master/src/Samples/Sample.DomainEvents.WebApi/Startup.cs#L98

However, when I use this registration method for my one sample producer/consumer, the following is logged:
```
[20:01:35 DBG] Found a base type of Namespace.XyzEvent that is configured in the bus: System.Object
```
Seems to me like the sample is missing the following piece of code down in the `.ForEach(find => ..)` method:
```
  builder.Produce(find.EventType,
                            x => x.DefaultTopic(x.Settings.MessageType.Name));
```

I'll happily contribute an PR fixing it if you agree :)

nover:
This pull request fixes the same code and documentation regarding the `InMemoryBus` and the auto-registration for producers and consumers.

The existing code failed to register the producers, leading to an unusable example for newcomers.

Closes #40

### DIFF
--- a/docs/provider_memory.md
+++ b/docs/provider_memory.md
@@ -88,6 +88,7 @@ mbb.Do(builder => applicationLayer.GetTypes().Where(t => t.IsClass && !t.IsAbstr
                     .ForEach(find =>
                     {
                         Log.InfoFormat(CultureInfo.InvariantCulture, "Registering {0} in the bus", find.EventType);
+                        builder.Produce(find.EventType, x => x.DefaultTopic(x.Settings.MessageType.Name));
                         builder.Consume(find.EventType, x => x.Topic(x.MessageType.Name).WithConsumer(find.HandlerType));
                     })
                 );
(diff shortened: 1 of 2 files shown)


nover:
@zarusz here you go, fixed samples :) Not quite sure what mr `DCO` is complaining about with my commits.

nover:
Just testing out the sample locally now, and if this line of code is kept:
```
                .Produce<OrderSubmittedEvent>(x => x.DefaultTopic(x.Settings.MessageType.Name))
```
In combination with my fix, then an exception is raised that `OrderSubmittedEvent` is declared multiple times.

zarusz:
Looks good @nover and thanks!

zarusz:
I will review that exception sometime soon and patch the runtime or the sample.

zarusz:
The `exception is raised that OrderSubmittedEvent is declared multiple times.`  should now be fixed with the referenced commit above.

Issue #155, opened 2023-03-09
Kafka Headers


correa-juan:
Hi ! 

We are sending an event with some headers but the headers are being sent  with extra quotes. 

![image](https://user-images.githubusercontent.com/127070539/223988964-f12a9bb9-5b74-453c-ba31-f7dd172fd6b9.png)

Do you know how to change that behavior ? 

thanks a lot ! 

Awesome project !

zarusz:
Hello, what version of SMB Kafka are you using? What version of serializer plugin? Do you setup custom kafka header serializer in SMB?

correa-juan:
Hello @zarusz  ! We are using: 

<img width="678" alt="image" src="https://user-images.githubusercontent.com/127070539/224062960-d9a67c56-832c-46be-b5ca-5be453094682.png">

this is our config: 

<img width="480" alt="image" src="https://user-images.githubusercontent.com/127070539/224063290-584394e3-d047-4e53-89fc-6fb3858f994f.png">

zarusz:
The underlyint client (Confluent.Kafka library) required the header values to come as `byte[]`.

So since you're using the JsonMessageSerializer here, values are serialized to JSON. In the example the primitive value is string, so it would be serialized to `"string here"` which is what is expected. It does look weird in the tool that you've using to view the headers in Kafka (what tool are you usung?).

Now, if you need to send a string representation in such case, then the following custom serializer should do the trick:

```cs
public class StringMessageSerializer : IMessageSerializer
{
    private readonly Encoding _encoding;

    public StringMessageSerializer(Encoding encoding = null)
    {
        _encoding = encoding ?? Encoding.UTF8;
    }

    #region Implementation of IMessageSerializer

    public byte[] Serialize(Type t, object message)
    {
        if (message == null) return null;
        var payload = _encoding.GetBytes(message.ToString());
        return payload;
    }

    public object Deserialize(Type t, byte[] payload)
    {
        if (payload == null) return null;
        var str = _encoding.GetString(payload);
        // ToDo: maybe int.TryParse if you need to detect the primitive type
        // ToDo: maybe Guid.TryParse
        // ToDo: maybe bool.TryParse 
        return str;
    }

    #endregion
}

mbb.    
   .WithProviderKafka(new KafkaMessageBusSettings(kafkaBrokers)
   {
      HeaderSerializer = new StringMessageSerializer()
   });
```

Let me know if this works for you.

[...]

PR #157, opened 2023-03-11
[Host.Kafka] Kafka default message serializer

zarusz:
It's better to have a simple serializer for Kafka header values that are set up by default regardless of the chosen message bus serializer.  The proposed serializer converts all the supplied values to string, and the UTF8 encoding is used by default to produce byte[] which the underlying Kafka client requires.

See #155

### DIFF
--- a/docs/provider_kafka.md
+++ b/docs/provider_kafka.md
@@ -125,7 +125,7 @@ SMB Kafka allows to set a provider (selector) that will assign the message key f
 
 ```cs
 // MessageBusBuilder mbb;
-mbb.    
+mbb    
    .Produce<MultiplyRequest>(x => 
    {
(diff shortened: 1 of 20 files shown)


zarusz:
@correa-juan I am in the process of working towards a 2.0.0 release of SMB. Check one of the latest 2.0.0-rcX releases or wait for a stable 2.0.0 release (now its a matter of a week).

Juandavi1:
we are trying it with: 
image

but the serializer cannot be imported 

image

it seems the rc version has not been published. 
The last rc is 4 days ago

zarusz:
@Juandavi1 the latest build is now released. Check 2.0.0-rc6 for Kafka.

PR #176, opened 2023-05-04
[Host.AzureServiceBus] Add option for replacing already defined rules in ASB subscription

ziurek:
This PR adds option `CanConsumerReplaceSubscriptionFilters` which is set to `false` by default. 
If it's `true`, then whenever any rules are already defined for ASB subscription, then SMB will replace them with newly defined ones.

Context: I observed an issue, if a subscription was defined without the rules, and after some time I decided to add some, then subscription had two rules:

- `$Default`
- and newly added rule

Expected behaviour is to remove the default one, once new rules are declared.

How to use this behaviour: set `ServiceBusTopologySettings`'s `CanConsumerReplaceSubscriptionFilters` flag to true.
Once you do it, if you defined any set of rules, they will override already exisiting ones.

NOTE: It compares already existing rules by name, so let say: if there is a rule with name `rule` and then you define new rule with name `rule` but different condition, it won't override it. To be discussed this should work like this, or rather it should compare condition as well.

Docs not yet updated.

Signed-off-by: Paweł Mazurek [ziurek0@gmail.com](mailto:ziurek0@gmail.com)

### DIFF
--- a/src/SlimMessageBus.Host.AzureServiceBus/ServiceBusTopologyService.cs
+++ b/src/SlimMessageBus.Host.AzureServiceBus/ServiceBusTopologyService.cs
@@ -38,6 +38,19 @@ public class ServiceBusTopologyService
         }
     }
 
+    private Task<Response> SwallowExceptionIfMessagingEntityNotFound(Func<Task<Response>> task)
+    {
+        try
+        {
+            return task();
+        }
+        catch (ServiceBusException e) when (e.Reason == ServiceBusFailureReason.MessagingEntityNotFound)
+        {
+            // do nothing as another service instance might have deleted entity in the meantime
+            return null;
+        }
+    }
+
     private Task<TopologyCreationStatus> TryCreateQueue(string path, bool canCreate, Action<CreateQueueOptions> action) => SwallowExceptionIfEntityExists(async () =>
     {
         if (await adminClient.QueueExistsAsync(path)) return TopologyCreationStatus.Exists;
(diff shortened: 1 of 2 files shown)


src/SlimMessageBus.Host.AzureServiceBus/ServiceBusTopologyService.cs
```diff
@@ -182,6 +184,19 @@ public async Task ProvisionTopology()
...
 
+                                    if (topologyProvisioning.CanConsumerReplaceSubscriptionFilters)
+                                    {
+                                        // Note: remove already defined rules for existing subscription
+                                        var removeRuleTasks = new List<Task<Response>>();
+                                        await foreach (var rulesPage in adminClient.GetRulesAsync(path, subscriptionName).AsPages())
+                                        {
+                                            removeRuleTasks
+                                                .AddRange(rulesPage.Values.Where(rule => !filters.Any(filter => filter.Name == rule.Name))
+                                                .Select(rule => adminClient.DeleteRuleAsync(path, subscriptionName, rule.Name)));
```
zarusz, line 195: When there are many concurrently running process instances, some of them might already delete the rule. 
Can you please wrap the DeleteRuleAsync into a try-catch in a similar way how the `SwallowExceptionIfEntityExists` is done?

ziurek: Ok, I've added this behaviour. Can you verify now?

src/SlimMessageBus.Host.AzureServiceBus/ServiceBusTopologyService.cs
```diff
@@ -38,6 +38,19 @@ private async Task<TopologyCreationStatus> SwallowExceptionIfEntityExists(Func<T
         }
     }
 
+    private Task<Response> SwallowExceptionIfMessagingEntityNotFound(Func<Task<Response>> task)
+    {
+        try
+        {
+            return task();
```
zarusz, line 45: I think we need to make this async, await to capture the ServiceBusException, otherwise it will be an AggregateTaskException.  Could you add the async/await keywords?

ziurek: Ok, I've updated the PR

Issue #223, opened 2024-03-29
[Host] CancellationToken is not included as an OnHandle method paramter


EtherZa:
Neither `IConsumer` nor `IRequestHandler` include the `CancellationToken` associated with the message context as a method parameter. 

It is accessible via `IConsumerContext` injection, but the implementation is not ideal as it breaks the facade of isolation from the message broker. Using setter injection also negates the ability to use a single consumer instance for concurrent processing.

zarusz:
I agree in general, it would be better to have CancellationToken added as part of the consumer method. 

There could also be an improvement added for concurrent consumers (`.Instances(N)` where `N > 1`) where the message processing pipeline would create N per-message scopes, rather than 1 nowadays.

In the long run, I was considering adding CancellationToken to the consumer methods in the future SMB v3. This perhaps would result in a breaking change to `IConsumer<T>`, or an additional interface that could be used.

I saw you have the PR for this feature using the [consumer without the interface](https://github.com/zarusz/SlimMessageBus/blob/master/docs/intro.md#consumer), which is a good middle-ground before v3.

zarusz:
Closed via #224

PR #224, opened 2024-03-29
[Host] Optionally include IConsumerContext and/or CancellationToken in consumer method invocation

EtherZa:
Added support for optionally including IConsumerContext and/or the CancellationToken in the consumer method invocation when supplying a type and method name. 

As the `CancellationToken` is tied to the life of the message lock, it should be readily available in the method to prematurely terminate processing when the lock is lost.

Supplying the `IConsumerContext` as an argument, as opposed to constructor or setting injection, allows for a single consumer instance where appropriate.

```cs
public class SomeConsumer
{
  public async Task MyHandleMethod(SomeMessage msg, IConsumerContext consumerContext, CancellationToken cancellationToken)
  {
    // handle the msg
  }
}
```

Please note that the order of the parameters is not important.

---
IConsumer and IRequestHandler have not been modified so as to avoid creating a breaking change. Implementation would be trivial should there be appetite to change the signatures.

Though not direct solutions to #220 and #223, the submission addresses both issues for non-IConsumer/IRequestHandler implementations.

### DIFF
--- a/src/SlimMessageBus.Host.Configuration/Builders/AbstractConsumerBuilder.cs
+++ b/src/SlimMessageBus.Host.Configuration/Builders/AbstractConsumerBuilder.cs
@@ -31,22 +31,53 @@ public abstract class AbstractConsumerBuilder : IAbstractConsumerBuilder
 
     static internal void SetupConsumerOnHandleMethod(IMessageTypeConsumerInvokerSettings invoker, string methodName = null)
     {
+        static bool ParameterMatch(IMessageTypeConsumerInvokerSettings invoker, MethodInfo methodInfo)
+        {
+            var parameters = new List<Type>(methodInfo.GetParameters().Select(x => x.ParameterType));
+
+            var requiredParameters = new[] { invoker.MessageType };
+            foreach (var parameter in requiredParameters)
+            {
+                if (!parameters.Remove(parameter))
+                {
+                    return false;
+                }
+            }
+
+            var allowedParameters = new[] { typeof(IConsumerContext), typeof(CancellationToken) };
+            foreach (var parameter in allowedParameters)
+            {
+                parameters.Remove(parameter);
+            }
+
+            if (parameters.Count != 0)
+            {
+                return false;
+            }
+
+            // ensure the method returns a Task or Task<T>
+            if (!typeof(Task).IsAssignableFrom(methodInfo.ReturnType))
+            {
+                return false;
+            }
+
+            return true;
+        }
+
         if (invoker == null) throw new ArgumentNullException(nameof(invoker));
 
         methodName ??= nameof(IConsumer<object>.OnHandle);
 
         /// See <see cref="IConsumer{TMessage}.OnHandle(TMessage)"/> and <see cref="IRequestHandler{TRequest, TResponse}.OnHandle(TRequest)"/> 
 
-        var consumerOnHandleMethod = invoker.ConsumerType.GetMethod(methodName, new[] { invoker.MessageType });
+        var consumerOnHandleMethod = invoker.ConsumerType.GetMethods(BindingFlags.Public | BindingFlags.Instance)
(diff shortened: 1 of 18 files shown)


EtherZa:
@zarusz there appears to be an issue with the SonarCloud secret

```
 The format of the analysis property sonar.token= is invalid
```

zarusz:
Yes, I believe secrets are not passed to branch builds which are from external repo clones. I will have a look when I get a chance.

zarusz:
@EtherZa I've changed the actions workflow some on master, can you please rebase your changes and push again?

src/SlimMessageBus.Host.Configuration/Builders/AbstractConsumerBuilder.cs
```diff
@@ -31,22 +33,53 @@ protected AbstractConsumerBuilder(MessageBusSettings settings, Type messageType,
 
     static internal void SetupConsumerOnHandleMethod(IMessageTypeConsumerInvokerSettings invoker, string methodName = null)
     {
+        bool ParameterMatch(MethodInfo methodInfo)
```
zarusz, line 36: Perhaps make this local func static

zarusz:
@EtherZa can you please rebase against master once more and ideally squash all your changes into 1 commit?
I am trying to tweak the build workflow so that SonarCloud and secrets for integration test infra are shared for PR coming from cloned repos. Unfortunately, that might require a couple of back and forth to get right.

I have looked at your changes - looks good!

src/SlimMessageBus.Host.Configuration/Builders/AbstractConsumerBuilder.cs
EtherZa: Done.

EtherZa:
@zarusz - No worries, I've rebased, squashed and pushed.

zarusz:
@EtherZa can you rebase once more?
I've pushed some more tweaks to build workflow.

I see that sonar cloud integrates nicely already.

After that I am happy to merge.

src/SlimMessageBus.Host/Helpers/ReflectionUtils.cs
```diff
@@ -40,6 +41,102 @@ public static T GenerateMethodCallToFunc<T>(MethodInfo method, Type instanceType
...
+    public static TDelegate GenerateMethodCallToFunc<TDelegate>(MethodInfo methodInfo, params Type[] argumentTypes)
+        where TDelegate : Delegate
+    {
+        if (methodInfo == null)
+        {
+            throw new ArgumentNullException(nameof(methodInfo));
+        }
+
+        var delegateSignature = typeof(TDelegate).GetMethod("Invoke")!;
+        var delegateReturn = delegateSignature.ReturnType;
```
zarusz, line 83: Yes, this variable seems not used.

EtherZa: Removed

EtherZa:
Rebased and pushed without the unused variable.

src/SlimMessageBus.Host.Configuration/Builders/AbstractConsumerBuilder.cs
```diff
@@ -31,22 +31,53 @@ protected AbstractConsumerBuilder(MessageBusSettings settings, Type messageType,
...
+            var allowedParameters = new[] { typeof(IConsumerContext), typeof(CancellationToken) };
+            foreach (var parameter in allowedParameters)
+            {
+                parameters.Remove(parameter);
+            }
+
+            if (parameters.Count != 0)
+            {
+                return false;
+            };
```
zarusz, line 56: Sorry, one last thing - the semicolon seems redundant

EtherZa:
Both SonarCloud issues resolved and pushed.

zarusz:
This is now merged. Thanks for your contribution @EtherZa! 
I will add this to the upcoming 2.3.0 release (soon).

EtherZa:
Thank you!

Issue #244, opened 2024-04-10
[Host.AzureServiceBus] Exception thrown when provisioning topolgy of a shared subscription


EtherZa:
When more than one consumer is configured to use the same subscription name with the Azure Service Bus transport, the `ServiceBusTopologyService` throws an exception when attempting to provision the subscription.

Sample configuration:

```cs
.Consume<AsbSample>(
    x => x.Topic("spike/slim")
        .SubscriptionName("one")
        .SubscriptionSqlFilter("CorrelationId = 'one'"))

.Consume<AsbSample2>(
    x => x.Topic("spike/slim")
        .SubscriptionName("one"))
```

Exception thrown in `ServiceBusTopologyService.ProvisionTopology()` as `x.SubscriptionName` is not unique.

![image](https://github.com/zarusz/SlimMessageBus/assets/1317506/c668a861-696d-400e-ad99-c149c5333482)

Messages are consumed, as expected, if the subscription already exists.

zarusz:
@EtherZa are you willing to adjust that and make the PR?

EtherZa:
> @EtherZa are you willing to adjust that and make the PR?

Yeah, sure. I'll see what I can whip up.

First thoughts:
1.  `ConsumerSettings` for a topic to be grouped by `SubscriptionName`.
2. All `CreateSubscriptionOptions` for the `SubscriptionName` need to be equal or throw `ConfigurationMessageBusException` if not.
3. Rules will be merged by name. Identical rules (name/filter/action) are fine; but rules with the same name but different filter criteria will throw `ConfigurationMessageBusException`.

zarusz:
Yes, the settings need to be the same otherwise it might not make sense for users to register two consumers on the same subscription.

zarusz:
Closed via #245

PR #245, opened 2024-04-11
#244: Topology provision for shared ASB subscription

EtherZa:
Resolution to exception thrown when attempting to provision topology with Azure Service Bus when multiple consumers share a single subscription.

1. `CreateSubscriptionOptions` across all consumers must be equal
2. Rules are grouped by name. Any grouped rules that are equal are treated as a single rule. Any variation on filters sharing the same name will result in aborting the topology provision.
3. @zarusz - I have added an additional `Updated` status to the `TopologyCreationStatus` enum to reflect success on updating a rule. Is this something that you favour, or should I stick with `Created`?  Do you want `Updated` added but with a value equal to `Created`?
4. An additional `ServiceBusTopologySettings` setting has been added - `CanConsumerValidateSubscriptionFilters`. The setting works as an extension to `CanConsumerCreateSubscriptionFilter` and `CanConsumerReplaceSubscriptionFilters` and allows for an extra gate to perform evaluation of the implemented subscription rules (server) and the code based subscription rules (no API calls will be made to ASB when all three are false).

Unit tests in `SlimMessageBus.Host.Test` have been set to run in series in order to avoid the race condition experienced with the library.

### DIFF
--- a/docs/provider_azure_servicebus.md
+++ b/docs/provider_azure_servicebus.md
@@ -444,6 +444,28 @@ This allows to establish ownership between services as to which one owns the top
 
 By default, all the flags are enabled (set to `true`). This is for convenience.
 
+In the case where multiple consumers share the same subscription name and topology provisioning is required, all `CreateConsumerOptions` must contain the same values for the same subscription. Filters are merged but must be equal if they use the same name.
+
+### Validation of Topology
+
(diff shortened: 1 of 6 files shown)


EtherZa:
@zarusz I've pushed the update and removed the WIP demarcation. It is on to you now :)

zarusz:
Awesome work @EtherZa!

Issue #254, opened 2024-04-26
$Default rule is not added to subscription


winiarski91:
After upgrading from version 2.2.3 to 2.3.3, we observed that the $Default rule is not being added to the subscription, leading to messages not being consumed from ASB.

zarusz:
Might be related to changes in #245.

EtherZa:
> Might be related to changes in #245.

It appears to be the case. #245 changed the flow to always delete the default rule regardless of whether there was a replacement rule or not.

PR #255 realigns the code base with the original workflow.

zarusz:
Thanks @EtherZa. I will have a look at your PR. Also started to look at it now - wanted to add more integration tests around cleaning of existing ASB topology in the tests, so that we can capture such in the future.

zarusz:
Fixed in version [2.3.4](https://www.nuget.org/packages/SlimMessageBus.Host.AzureServiceBus/2.3.4)

PR #255, opened 2024-04-28
#254: Do not delete default rule if no other rule exists

EtherZa:
Bug in #245 resulted in the default rule of a subscription being deleted on creation regardless as to where there was a rule to replace it or not.

PR returns to the original flow where the default rule will only be deleted if the subscription has just been created and there are alternate rules to apply.


zarusz:
@EtherZa I think some more checks are missing further down and the int test is failing. However, I've also made progress on my own fix in #257. Can we use that other PR for the resolution of the bug, please? There is more extras I've added on the way.

PR #257, opened 2024-04-28
[Host.AzureServiceBus] $Default rule is not added to subscription #254

### LINKED ISSUE #254: $Default rule is not added to subscription

zarusz:
Fixes #254

### DIFF
--- a/src/Host.Plugin.Properties.xml
+++ b/src/Host.Plugin.Properties.xml
@@ -4,7 +4,7 @@
 	<Import Project="Common.NuGet.Properties.xml" />
 
 	<PropertyGroup>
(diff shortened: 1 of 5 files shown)


zarusz:
@EtherZa do you mind reviewing this change?
* I have added the ability to intercept ASB topology change and be able to run cleanup before tests
* Refactored the checking for declared subscription rules against server riles and taking `$Default` into account

EtherZa (approved):
Looks good.

Issue #260, opened 2024-06-12
[Host.Outbox | Host.Outbox.Sql] Poor performance when publishing outbox messages


EtherZa:
The performance of the outbox plug-in can be greatly improved; especially when publishing messages from the outbox/db.

A few of the issues:

SQL:
1. PK of outbox table is using a clustered index with a unique identifier
2. Locking of the table updates all undelivered rows instead of just those included in the batch
4. Updating of delivery status is by multiple statements instead of a table-valued parameter (TVPs)
5. Only a single application instance can spool messages (always FIFO, but not always required).

Host:
1. Interceptors are executed on initial publish (before outbox) and again when being submitted to message bus (after outbox)
2. Messages are trickle fed from the outbox to the message bus even when providers support batches

---
Proposal:
1. Improve SQL schema, indexes and queries
2. Apply an order to the outbox interceptor so that it is always the last interceptor to run (and skip running interceptors when publishing from outbox)
3. Modify `MessageBusBase.ProduceToTransport` to accept batches of messages for the same bus and path
4. Modify `OutboxSendingTask` to group messages for the same bus and path together and submit to `MessageBusBase.ProduceToTransport` directly without having to traverse through the pipeline stack again.
5. Submit batches of messages to buses that support it while continuing to trickle feed to those that don't

zarusz:
I am prepping a quick benchmark test to be able to assess the gains of this PR.
Once I have something, I will report here.

EtherZa:
I'm busy refactoring the `OutboxSenfingTask` to drop support for non-`IMessageBulkProducer` providers as you suggested.

As part of it, I'm also changing the lock to renew on a timer instead of killing the batch process. This should provide a little more safety when handling issues like ASB throttling.

I have also uncovered a potential critical issue which I'm putting a fix in for -the `DeliveryAttempt` value is set, but never evaluated. If a message ended up being too big for the queue to accept (for example), the current implementation will just keep trying indefinitely. Apart from the obvious resourcing problem, it would eventually lead to all available 'slots' in the batch being taken by such failures and the outbox stalling. I am just going to limit the attempts without delays, etc for now as it is an edge case. More can come later.

zarusz:
@EtherZa the baseline test + 2 more lighter enhancements is in #263. Feel free to comment. 
Ideally, I would merge this to master, and if you could rebase your changes after.

[...]

PR #261, opened 2024-06-12
[Host | Host.Outbox | Host.Outbox.Sql] #260 - Sql outbox performance improvments (batching + optimised queries)

EtherZa:
This comes as (another) unsolicited/undiscussed PR after investigating some performance issues that I was experiencing. Please feel free to be absolutely brutal with it and/or any aspect that is not to your liking.

The most jarring change is probably going to be with the addition of `IMessageBusBulkProducer` and the required changes to `MessageBusBase.ProduceToTransport` to allow batching and access from `OutboxSendingTask`.

Changes:
1. SQL far more performant. Index improvements, limiting soft locks to fewer rows and single queries/TVP's have had a profound impact. In testing, I was seeing no performance degradation as the table grew to several million entries.
3. `OutboxSettings.MaintainSequence = false` will allow for load balanced instances to process the outbox together (default is true to keep same behaviour).
4. `MessageBusBase.ProduceToTransport` signature has been changed to accept multiple messages for the same bus/path. Serialization has also moved into the method to allow the same.
5.  `OutboxForwardingPublishInterceptor` is ordered with max int to ensure that it is the last interceptor to run. Publishing of messages from the outbox do not run through the interceptor pipeline (already done so before being placed in the outbox).
6. Batching support has been added to the Azure Service Bus and Azure Event Hub providers. Publishing of messages is in multiple orders of magnitude quicker than before (not benchmarked). Messages are submitted in batches of up to 100 by default.

The submission is lacking in additional unit tests at the moment (can be added once direction is determined), but existing tests should cover most scenarios as the flow should not change. One test was removed as it was no longer relevant.

---

Again, I completely understand that this is undiscussed. It was thrown together while looking at what could be done to improve the implementation, so please feel free to fob it off as heading in the wrong direction or give any changes that you feel would provide something more to your liking.
-Richard


zarusz:
Thanks for looking into this.

Yes, I believe the outbox could have deserved a performance Improvement pass since it's inception.

I wanted to improve the performance of outbox especially when going into millions of messages and database contention in another project that SMB is used we saw some potential for improvement. So this should align with that I believe.

The batch send was also something on the roadmap, so it's another good avenue. 

Let me take the time to review it in depth and align the implementation. 

I would also love to add some performance test to he able to baseline before and after.

src/SlimMessageBus.Host/MessageBusBase.cs
```diff
@@ -408,7 +410,22 @@ protected virtual string GetDefaultPath(Type messageType, ProducerSettings produ
         return path;
     }
 
-    protected abstract Task ProduceToTransport(object message, string path, byte[] messagePayload, IDictionary<string, object> messageHeaders, IMessageBusTarget targetBus, CancellationToken cancellationToken = default);
+    protected async Task ProduceToTransport(object message, Type messageType, string path, IDictionary<string, object> messageHeaders, IMessageBusTarget targetBus, CancellationToken cancellationToken = default)
+    {
+        var envelope = new Envelope(message, messageType, messageHeaders);
+        var result = await ProduceToTransport([envelope], path, targetBus, cancellationToken);
```
zarusz, line 416: Just a thought:
We end up adding 2 more allocations overhead for every msg publish (that is not part of the outbox route).

src/SlimMessageBus.Host.AzureEventHub/EventHubMessageBus.cs
```diff
@@ -101,42 +103,105 @@ protected override async Task OnStart()
...
+                            throw new ApplicationException($"Failed to add message {item.Envelope.Message} of Type {item.Envelope.MessageType?.Name} on Path {path} to an empty batch");
+                        }
+
+                        inBatch.Add(item.Envelope);
+                    }
+
+                    if (batch != null)
+                    {
+                        await producer.SendAsync(batch, cancellationToken).ConfigureAwait(false);
+                        _logger.LogDebug("Batch of {BatchSize} message(s) dispatched to {path} ({SizeInBytes} bytes)", batch.Count, path, batch.SizeInBytes);
```
zarusz, line 185: {Path}

src/SlimMessageBus.Host.AzureEventHub/EventHubMessageBus.cs
```diff
@@ -101,42 +103,105 @@ protected override async Task OnStart()
...
+                        var item = items.Single();
+                        await producer.SendAsync([item.Message], new SendEventOptions { PartitionKey = partition.Key }, cancellationToken);
+
+                        dispatched.Add(item.Envelope);
+                        continue;
+                    }
+
+                    // multiple items - send in batches
+                    var inBatch = new List<T>(items.Count);
+                    foreach (var item in items)
```
zarusz, line 159: To avoid repetition of CreateBatchAsync and TryAdd within the loop I would replace the foreach loop with just an loop that iterates over items with an index `i` and does not advance to `i++` if the TryAdd failed.
Also for the last element it could also invike SendAsync for  the batch so that we don't have to call it again outside of the loop.

Something like this (did not check corectness):

```cs
for(var i = 0;  i < items.Count; ) 
{
   var item = items[i];
batch ??= await producer.CreateBatchAsync(new CreateBatchOptions { PartitionKey = partition.Key }, cancellationToken);
                        if (batch.TryAdd(item.Message))
                        {
                            inBatch.Add(item.Envelope);
                            i++;
                            // be sure to send if last message
                             if (i < items.Count) {
                                                       continue;
                            } 
                        }

                        await producer.SendAsync(batch, cancellationToken).ConfigureAwait(false);
                        dispatched.AddRange(inBatch);
                        inBatch.Clear();
                        batch.Dispose();
}
```

src/SlimMessageBus.Host.AzureServiceBus/ServiceBusMessageBus.cs
```diff
@@ -113,48 +116,114 @@ void AddConsumerFrom(TopicSubscriptionParams topicSubscription, IMessageProcesso
...
-        if (messageType != null)
-        {
-            // local producer modifier second
-            var producerSettings = GetProducerSettings(messageType);
-            InvokeMessageModifier(message, messageType, m, producerSettings);
+                return ([item.Envelope], null);
+            }
+            catch (Exception ex)
+            {
+                _logger.LogDebug(ex, "Producing item {Message} of type {MessageType} to path {Path} resulted in error {Error}", item.Envelope.Message, item.Envelope.MessageType?.Name, path, ex.Message);
```
zarusz, line 172: I like the previous `Producing *message*` better than the `Producing *item*` terminology

src/SlimMessageBus.Host.AzureServiceBus/ServiceBusMessageBus.cs
```diff
@@ -113,48 +116,114 @@ void AddConsumerFrom(TopicSubscriptionParams topicSubscription, IMessageProcesso
...
-        var senderClient = _producerByPath.GetOrAdd(path);
-
+        // multiple items - send in batches
+        var dispatched = new List<T>(envelopes.Count);
+        ServiceBusMessageBatch batch = null;
         try
         {
-            await EnsureInitFinished();
+            var inBatch = new List<T>(envelopes.Count);
+            foreach (var item in messages)
```
zarusz, line 183: I'd also think the other loop iteration approach as I explained for EventHub could work better here (less repeated code)

src/SlimMessageBus.Host.AzureServiceBus/ServiceBusMessageBus.cs
```diff
@@ -167,7 +236,7 @@ private void InvokeMessageModifier(object message, Type messageType, ServiceBusM
         }
         catch (Exception e)
         {
-            _logger.LogWarning(e, "The configured message modifier failed for message type {MessageType} and message {Message}", messageType, message);
+            _logger.LogWarning(e, "The configured item modifier failed for item type {MessageType} and item {Message}", messageType, message);
```
zarusz, line 239: Message term should be better than item term.

src/SlimMessageBus.Host.Redis/RedisMessageBus.cs
```diff
@@ -147,32 +147,51 @@ void AddTopicConsumer(string topic, ISubscriber subscriber, IMessageProcessor<Me
 
     #region Overrides of MessageBusBase
 
-    protected override async Task ProduceToTransport(object message, string path, byte[] messagePayload, IDictionary<string, object> messageHeaders, IMessageBusTarget targetBus, CancellationToken cancellationToken = default)
+    protected override async Task<(IReadOnlyCollection<T> Dispatched, Exception Exception)> ProduceToTransport<T>(IReadOnlyCollection<T> envelopes, string path, IMessageBusTarget targetBus, CancellationToken cancellationToken = default)
     {
-        if (message is null) throw new ArgumentNullException(nameof(message));
-        if (messagePayload is null) throw new ArgumentNullException(nameof(messagePayload));
+        if (envelopes is null) throw new ArgumentNullException(nameof(envelopes));
+        if (envelopes.Any(x => x.Message is null))
```
zarusz, line 153: I think checking every message here might be an overkill. This should never be null.

src/SlimMessageBus.Host.Sql.Common/CommonSqlMigrationService.cs
```diff
@@ -113,7 +113,7 @@ public async virtual Task Migrate(CancellationToken token)
         }
         catch (SqlException e)
         {
-            Logger.LogError(e, "Database schema provisioning enocuntered a non-recoverable SQL error: {ErrorMessage}", e.Message);
+            Logger.LogError(e, "Database schema provisioning encountered a non-recoverable SQL error: {ErrorMessage}", e.Message);
```
zarusz, line 116: Good job capturing typos!

src/SlimMessageBus.Host/IMessageBusBulkProducer.cs
```diff
@@ -0,0 +1,22 @@
+﻿namespace SlimMessageBus.Host
+{
+    using System.Collections.Generic;
```
zarusz, line 3: Implicit usings please

src/SlimMessageBus.Host/IMessageBusBulkProducer.cs
```diff
@@ -0,0 +1,22 @@
...
+    /// <summary>
+    /// Interface for bulk publishing of messages directly to a queue.
+    /// </summary>
+    /// <remarks>
+    /// The Interface is intended for outbox publishing where messages have already flowed through the interceptor pipeline.
+    /// </remarks>
+    public interface IMessageBusBulkProducer
+    {
+        /// <summary>
+        /// The maximum number of messages that cen take part in a <see cref="TransactionScope"/>. Null if transaction scopes are not supported by the message bus.
```
zarusz, line 15: Typo `cen` :)

src/Tests/SlimMessageBus.Host.Test/MessageBusBaseTests.cs
```diff
@@ -245,42 +245,6 @@ private static async Task WaitForTasks(int millis, params Task[] tasks)
         }
     }
 
-    [Fact]
-    public async Task When_Produce_DerivedMessage_Given_OnlyBaseMessageConfigured_Then_BaseMessageTypeIsSentToSerializer()
```
zarusz, line 249: Not following why this test had to be removed. Am I missing something?

src/SlimMessageBus.Host.Outbox.Sql/SqlOutboxTemplate.cs
```diff
@@ -4,24 +4,101 @@ public class SqlOutboxTemplate
...
+        SqlOutboxMessageLockAndSelect = $"""
+            WITH Batch AS (SELECT TOP (@BatchSize) *
+                           FROM {TableNameQualified}
+                           WHERE DeliveryComplete = 0
+                             AND (LockInstanceId = @InstanceId
+                               OR LockExpiresOn < GETUTCDATE())
+                           ORDER BY Timestamp ASC),
+                 SortedBatch AS (SELECT TOP (@BatchSize) *
+                                 FROM Batch
+                                 ORDER BY BusName, path, Timestamp)
```
zarusz, line 39: Path

src/SlimMessageBus.Host.Outbox.Sql/SqlOutboxMigrationService.cs
```diff
@@ -50,5 +50,30 @@ protected override async Task OnMigrate(CancellationToken token)
...
+            -- drop old indexes
+            DROP INDEX IX_Outbox_InstanceId ON {qualifiedTableName};
+            DROP INDEX IX_Outbox_LockExpiresOn ON {qualifiedTableName};
+            DROP INDEX IX_Outbox_Timestamp_LockInstanceId ON {qualifiedTableName};
+
+            -- SqlOutboxTemplate.SqlOutboxMessageLockAndSelect
+            CREATE INDEX IX_Outbox_Timestamp_LockInstanceId_LockExpiresOn_DeliveryComplete_0 ON {qualifiedTableName} (Timestamp, LockInstanceId, LockExpiresOn) WHERE (DeliveryComplete = 0);
+
+            -- SqlOutboxTemplate.SqlOutboxMessageLockTableAndSelect
+            CREATE INDEX IX_Outbox_LockExpiresOn_LockInstanceId_DeliveryComplete_0 ON {qualifiedTableName} (LockExpiresOn, LockInstanceId) WHERE DeliveryComplete = 0
```
zarusz, line 69: For consistendy add ( ) in DeliveryComplete = 0

src/SlimMessageBus.Host.Outbox/Interceptors/OutboxSendingTask.cs
```diff
@@ -166,56 +144,103 @@ private async Task Run()
...
 
-    private async Task<bool> SendMessages(IServiceProvider serviceProvider, IOutboxRepository outboxRepository, List<Guid> processedIds, CancellationToken ct)
+    private async Task SendMessagesAsync(IServiceProvider serviceProvider, IOutboxRepository outboxRepository, CancellationToken cancellationToken)
     {
+        async Task DispatchAndResetBatchAsync(IMessageBusBulkProducer producer, IMessageBusTarget messageBusTarget, IList<EnvelopeWithId> batch, string busName, string path)
+        {
+            _logger.LogDebug("Publishing batch of {MessageCount} messages to path {Path} on {BusName} bus", batch.Count, path, busName);
+
+            // TOOD: Enclose in a transaction
+            await producer.ProduceToTransport([.. batch], path, messageBusTarget, cancellationToken).ConfigureAwait(false);
```
zarusz, line 154: Would it make more sense to pass the original collection `batch` and not to create a collection copy?

src/SlimMessageBus.Host.Outbox/Interceptors/OutboxSendingTask.cs
```diff
@@ -166,56 +144,103 @@ private async Task Run()
...
+            await outboxRepository.UpdateToSent([.. batch.Select(x => x.Id)], cancellationToken).ConfigureAwait(false);
```
zarusz, line 155: Shouldn't this get the list of dispatched messages (result from ProduceToTransport) and only mark these as sent?

src/SlimMessageBus.Host.Outbox/Interceptors/OutboxSendingTask.cs
```diff
@@ -166,56 +144,103 @@ private async Task Run()
...
-                        _logger.LogDebug("Stopping the outbox message processing after {MessageCount} (out of {BatchCount}) because the message lock was close to expiration {LockBuffer}", i, _outboxSettings.PollBatchSize, _outboxSettings.LockExpirationBuffer);
-                        hasMore = false;
-                        break;
-                    }
-
-                    var bus = GetBus(compositeMessageBus, messageBusTarget, outboxMessage.BusName);
-                    if (bus == null)
+                    i++;
+                    var now = DateTime.UtcNow;
+                    if (DateTime.UtcNow.Add(_outboxSettings.LockExpirationBuffer) > outboxMessage.LockExpiresOn)
```
zarusz, line 181: Shouldn't the `now` be used instead of `DateTime.UtcNow` in here?

src/SlimMessageBus.Host.Outbox/Interceptors/OutboxSendingTask.cs
```diff
@@ -166,56 +144,103 @@ private async Task Run()
...
+                    if (bulkProducer == null)
                     {
-                        await bus.ProducePublish(message, path: outboxMessage.Path, headers: headers, messageBusTarget, cancellationToken: ct);
+                        // Add special header to supress from forwarding the message again to outbox
+                        headers.Add(OutboxForwardingPublishInterceptor<object>.SkipOutboxHeader, string.Empty);
 
-                        processedIds.Add(outboxMessage.Id);
+                        if (!cancellationToken.IsCancellationRequested)
+                        {
+                            await bus.ProducePublish(message, path: outboxMessage.Path, headers: headers, messageBusTarget, cancellationToken: cancellationToken);
```
zarusz, line 227: It needs to be noted in the Outbox documentation, that for transports that support bulk produce the interceptors won't be called for dispatching of an msg that was hydrated from outbox.

src/SlimMessageBus.Host.Outbox/Interceptors/OutboxSendingTask.cs
```diff
@@ -166,56 +144,103 @@ private async Task Run()
...
     {
+        async Task DispatchAndResetBatchAsync(IMessageBusBulkProducer producer, IMessageBusTarget messageBusTarget, IList<EnvelopeWithId> batch, string busName, string path)
+        {
+            _logger.LogDebug("Publishing batch of {MessageCount} messages to path {Path} on {BusName} bus", batch.Count, path, busName);
+
+            // TOOD: Enclose in a transaction
+            await producer.ProduceToTransport([.. batch], path, messageBusTarget, cancellationToken).ConfigureAwait(false);
+            await outboxRepository.UpdateToSent([.. batch.Select(x => x.Id)], cancellationToken).ConfigureAwait(false);
+
+            batch.Clear();
```
zarusz, line 157: The messages that got dispatched have to be removed.
The message that failed has to have the delivery attempt increased.

src/SlimMessageBus.Host.Outbox/Interceptors/OutboxSendingTask.cs
```diff
@@ -166,56 +144,103 @@ private async Task Run()
...
+
                     _logger.LogDebug("Sending outbox message with Id {MessageId} of type {MessageType} to path {Path} using {BusName} bus", outboxMessage.Id, outboxMessage.MessageType.Name, outboxMessage.Path, outboxMessage.BusName);
                     var message = bus.Serializer.Deserialize(outboxMessage.MessageType, outboxMessage.MessagePayload);
-
-                    // Add special header to supress from forwarding the message againt to outbox
                     var headers = outboxMessage.Headers ?? new Dictionary<string, object>();
-                    headers.Add(OutboxForwardingPublishInterceptor<object>.SkipOutboxHeader, string.Empty);
 
-                    if (!ct.IsCancellationRequested)
+                    if (bulkProducer == null)
```
zarusz, line 220: I think all of the bus transports will implement the `IMessageBulkProducer` - they all inherit from MessageBusBase.
Perhaps its worth removing this case and adjusting `OutboxForwardingPublishInterceptor` to never check `SkipOutboxHeader`.

zarusz:
@EtherZa  you're definitely on the right track. I've reviewed the PR and provided some suggestions to consider.

As a side note, the bulk produce interface might be leveraged down the road in v3 to be able to send a batch of messages, or when (future releases?) we will bring in an auto batching of single message publishes.

I would love to see some benchmark test (as an integration test) using the existing Outbox infrastructure on Azure. I can take that on in a second iteration of this feature.

src/SlimMessageBus.Host.AzureServiceBus/ServiceBusMessageBus.cs
EtherZa: I'm not sure how that slipped in. Thank you!

src/SlimMessageBus.Host.Redis/RedisMessageBus.cs
EtherZa: It was based on the previous implementation, but I agree - removed.

src/SlimMessageBus.Host/IMessageBusBulkProducer.cs
EtherZa: Good spot!

src/SlimMessageBus.Host.Outbox/Interceptors/OutboxSendingTask.cs
EtherZa: Absolutely.

src/SlimMessageBus.Host.Outbox/Interceptors/OutboxSendingTask.cs
EtherZa: Ha! You found the 2am change. Did all the work and then didn't make use of it! Thank you :)

src/SlimMessageBus.Host.Outbox/Interceptors/OutboxSendingTask.cs
EtherZa: Locking is done in UTC and applied by script. As there is a chance that the SQL server and application may be in different time zones, this will keep it in sync.

src/SlimMessageBus.Host.Outbox.Sql/SqlOutboxTemplate.cs
```diff
@@ -4,24 +4,101 @@ public class SqlOutboxTemplate
...
+            FROM SortedBatch
+            ORDER BY BusName, path;
+            """;
 
-        SqlOutboxMessageFindNextSelect = @$"SELECT TOP {settings.PollBatchSize} * FROM {TableNameQualified} WHERE DeliveryComplete = 0 AND LockInstanceId = @InstanceId ORDER BY Timestamp ASC";
-        SqlOutboxMessageTryLockUpdate = @$"UPDATE {TableNameQualified} SET LockInstanceId = @InstanceId, LockExpiresOn = @ExpiresOn WHERE DeliveryComplete = 0 AND LockExpiresOn < GetUtcDate() OR LockInstanceId = @InstanceId";
-        SqlOutboxMessageDeleteSent = @$"DELETE FROM {TableNameQualified} WHERE [DeliveryComplete] = 1 AND [Timestamp] < @Timestamp";
+        SqlOutboxMessageUpdateSent = $"""
+            UPDATE {TableNameQualified}
+            SET [DeliveryComplete] = 1 WHERE [Id] IN (SELECT [Id] from @Ids)
```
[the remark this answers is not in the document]
EtherZa: @Ids is a table-valued parameter.

src/SlimMessageBus.Host.Outbox/Interceptors/OutboxSendingTask.cs
EtherZa: The only one that isn't right now is the `HybridMessageBus`. It probably could implement `IMessageBulkProducer` but i haven't looked at it TBH. I thought that as there could potentially be third party providers that won't implement it, it should be there as a fallback -but if you feel differently, please let me know.

src/Tests/SlimMessageBus.Host.Test/MessageBusBaseTests.cs
EtherZa: The test verified that the message was serialized before the call to `ProduceToTransport`. Serialization now happens in `ProduceToTransport` as part of the message generation (and as part of the specialized providers responsibility).

The test verified the serialization by mocking `ProduceToTransport`.

src/SlimMessageBus.Host/MessageBusBase.cs
EtherZa: Yeah, it's not great, but I can't think of an alternative implementation that does not involve creating two versions of `ProduceToTransport` for each provider. I eventually waived it away as the only path to this is by sending a single message where the hit should be insignificant when combined with the network latency of actually sending the message (memory transport excluded -which does not implement it anyway).

If you have a better idea, please let me know :)

EtherZa:
> I would also love to add some performance test to he able to baseline before and after.

I hear you with this, but TBH, I don't know how/what to benchmark. The gains are primarily through reducing the number of calls over the network and the cumulative latency between those calls. Mocking out the calls would nullify the benchmark and leaving them in puts the weighting on latency which is dependent on available bandwidth and round trip time.

EtherZa:
> I would love to see some benchmark test (as an integration test) using the existing Outbox infrastructure on Azure. I can take that on in a second iteration of this feature.

For my own testing, I used the `Given_CommandHandlerInTransaction_When_ExceptionThrownDuringHandlingRaisedAtTheEnd_Then_TransactionIsRolledBack_And_NoDataSaved_And_NoEventRaised` test and manually dumping several thousand messages into a local containerised db. The difference is chalk and cheese.

EtherZa:
> As a side note, the bulk produce interface might be leveraged down the road in v3 to be able to send a batch of messages, or when (future releases?) we will bring in an auto batching of single message publishes.

Great stuff. There are further improvements that could be made with the outbox model too. 
- Batching a large number of outbox bound messages through a TVP.
- The `OutboxSendingTask` can vastly increase the polling interval if the `OutboxForwardingPublishInterceptor` were to emit an event after the "outboxing" transaction completes. Each balanced app instance could immediately pick up that it had placed a message in the outbox, with others checking in periodically. This could be extended with support of asynchronous notifications à la postgres `notify`.
- The instance lock should renew itself instead of stopping to process messages when the time enters the `LockExpirationBuffer` window. The `LockExpirationBuffer` approach is risky if one part of the operation (publish or SQL sent status) is partially complete.

Battles need to be picked though :)

src/SlimMessageBus.Host.Outbox/Interceptors/OutboxSendingTask.cs
EtherZa: Need to read comments a little closer. Yes, use the now *variable* - done.

src/SlimMessageBus.Host.Outbox/Interceptors/OutboxSendingTask.cs
zarusz: The `GetBus()` will always lookup the child (target) bus when hybrid is used.
If the hybrid bus is not used then it will grab the chosen bus transport.

In either case all of these transport will always extend the MessageBusBase which implements this transport.

I think it's safe to assume this will be true for all transports to come in the future, and therefore it would remove some code from here.

src/SlimMessageBus.Host/MessageBusBase.cs
zarusz: I was also thinking about having 2 implementations in that case, but that creates more code and not sure about the benefit vs overhead if it truly warrants going down that route.

Also, with these changes some transports do not truly send the messages in bulk, and maybe letting the interface `IMessageBusBulkProducer` be implemented only by the transports that support bulk. 

I am a bit on the fence right now. 
However, let's proceed and perhaps I will refactor this a bit in the upcoming time.

src/SlimMessageBus.Host.Outbox/Interceptors/OutboxSendingTask.cs
EtherZa: Will do. The entire loop can be refactored into something simpler. It was originally written to make use of an `IAsyncEnumerable`, but that fell away with `SqlOutboxRepository` using a single connection.

EtherZa:
I have rebased from master and the run the first set of benchmarking tests that @zarusz added -thank you for adding them.

The results are skewed as the outbox is unfairly advantaged by having vastly lower latency and higher bandwidth available to it than the ASB, but do show an improvement.

Each set  od results are from three consecutive runs on a release build using a local SQL instance (containerised) with an Azure Service Bus (Standard tier) in the Australia East region. I have +- 78ms latency with the Australia East data centre (https://www.azurespeed.com/Azure/Latency).

### master
```
direct
Message Publish took	: 00:00:05.1472734  00:00:04.8023864  00:00:04.8906686
Message Consume took	: 00:00:08.2347485  00:00:07.9732388  00:00:07.1775174

with outbox
Message Publish took	: 00:00:01.1531716  00:00:00.7929830  00:00:00.7882242
Outbox Publish took	: 00:01:14.1196937  00:01:14.7346352  00:01:15.8530969
Message Consume took	: 00:00:06.9526368  00:00:06.3923101  00:00:06.6098189
```
### feature/260-sql-outbox-batching
```
direct
Message Publish took	: 00:00:04.8164905  00:00:04.6085346  00:00:05.1897898
Message Consume took	: 00:00:06.8671870  00:00:06.6343886  00:00:07.3392817

with outbox
Message Publish took	: 00:00:01.3155528  00:00:00.7840798  00:00:00.8276036
Outbox Publish took	: 00:00:04.0035331  00:00:04.6453286  00:00:04.5108678
Message Consume took	: 00:00:06.6543769  00:00:06.9201661  00:00:06.8933468
```

There are more changes needed before this will be ready to merge

1. Performance of non-bulk transports will be poor at the moment (worse than original)
2. Although failed delivery attempts are recorded, the current implementation does not process those that fail any differently to a new message (if a failure is due to size/incompatibility with the svb, the outbox will eventually stagnate)
3. The locking mechanism is not optimal. Instead of aborting progress when approaching the end of lock time, it would be better to extend the lock.

I am busy implementing the change to the three items above and will push asap.

Out of interest, I also ran single batches of 10k and 100k messages for the "with outbox" test only.

### 10k
```
Message Publish took	: 00:00:08.2749961
Outbox Publish took	: 00:00:18.5741174
Message Consume took	: 00:01:07.0192878
```

### 100k
```
Message Publish took	: 00:01:20.7375655
Outbox Publish took	: 00:02:59.2045206
Message Consume took	: 00:11:00.8795735
```

zarusz:
@EtherZa It is exciting to see way better numbers after your changes, good job!

A couple of thoughts:
- It would be great to work on additional identified improvements (1-3) - perhaps in future PRs.
It might be good to capture the numbers (from your local and/or from the github actions) and to make a note on the Outbox docs - so that folks have a feel about the overhead.
- I might consider refactoring and segmenting the bulk vs non-bulk capable transports (low on my priority list). In the future (perhaps v3) elevating the core interfaces to be able to produce a batch of messages which would go hand in hand with this PR.
- Looking at the consumer times for 10k v 100k for ASB it would require adding partitioning and more concurrent consumers to let it scale better.

EtherZa:
@zarusz I have pushed a commit for three items above. Please let me know if you are happy with the change, and I will rebase and sign. If you would prefer to defer it to another PR in order to limit scope/provide an alternate solution that's fine too -I can remove the commit easily from here.

Latest change includes:
* Drops support for buses that do not implement `IMessageBusBulkProducer`. `OutboxSendingTask` can now be understood at a glance :)
* Renew locks instead of aborting when time is nearly done (`LockExpirationBuffer`)
* Apply a max number of delivery attempts before aborting a message. This does need further work as there are no back off strategies, etc; but at least it will not block the queue now. Indexes/Queries have been updated and tested against a table of 10m with no degradation in performance.
* Azure Service Bus will now back off for 2.x seconds as per `ServiceBusException` (`ServiceBusFailureReason.ServiceBusy`). This should probably sit behind a gate to ensure all processes adhere to the cool off, but for now only the current execution path is affected.

> It might be good to capture the numbers (from your local and/or from the github actions) and to make a note on the Outbox docs - so that folks have a feel about the overhead.

I think my timings are probably a little too biased. While there is a substantial performance improvement, mine are definitely stacked against the real world with the database being local and the svb being remote. GitHub actions would be better as there is no favour and they are repeatable/independent ...the only problem is that the timings are not being published (no logs for successful tests).

> I might consider refactoring and segmenting the bulk vs non-bulk capable transports (low on my priority list). In the future (perhaps v3) elevating the core interfaces to be able to produce a batch of messages which would go hand in hand with this PR.

Supporting batch uploads to SVB/SQL would be a nice to have and would definitely be useful but probably not used too often.

In terms of performance, the impact from the boxing looks negligible IMO. It is too closely aligned with IO to make any tangible difference.
```
Before - Message Publish took	: 00:00:05.1472734  00:00:04.8023864  00:00:04.8906686
After  - Message Publish took	: 00:00:04.8164905  00:00:04.6085346  00:00:05.1897898
```

> Looking at the consumer times for 10k v 100k for ASB it would require adding partitioning and more concurrent consumers to let it scale better.

As long as a single service can publish more messages from the outbox than that service can consume, I can't see an issue. It could become a problem if order is important though. It would kill scaling out in its current form. In that case, it would be better to lock by bus/path or custom strategy, so that multiple instances could each deliver for different streams.

Please let me know your thoughts and I'll take it from there.

src/SlimMessageBus.Host.AzureServiceBus/ServiceBusMessageBus.cs
```diff
@@ -113,48 +118,139 @@ void AddConsumerFrom(TopicSubscriptionParams topicSubscription, IMessageProcesso
         }
     }
 
-    protected override async Task ProduceToTransport(object message, string path, byte[] messagePayload, IDictionary<string, object> messageHeaders, IMessageBusTarget targetBus, CancellationToken cancellationToken = default)
+    protected override async Task<(IReadOnlyCollection<T> Dispatched, Exception Exception)> ProduceToTransport<T>(IReadOnlyCollection<T> envelopes, string path, IMessageBusTarget targetBus, CancellationToken cancellationToken = default)
     {
-        var messageType = message?.GetType();
+        async Task SendBatchAsync(ServiceBusSender senderClient, ServiceBusMessageBatch batch, CancellationToken cancellationToken)
+        {
+            var attempts = 0;
```
zarusz, line 125: You could extract this retry login into a util method in the `SlimMessageBus.Host` project.
This might be reused in a few places, and increase the code readability here.

src/SlimMessageBus.Host.AzureServiceBus/ServiceBusMessageBus.cs
```diff
@@ -113,48 +118,139 @@ void AddConsumerFrom(TopicSubscriptionParams topicSubscription, IMessageProcesso
...
+                await SendBatchAsync(senderClient, batch, cancellationToken).ConfigureAwait(false);
+
+                dispatched.AddRange(inBatch);
+                inBatch.Clear();
+                batch.Dispose();
+                batch = null;
+            }
 
-            await senderClient.SendMessageAsync(m, cancellationToken: cancellationToken).ConfigureAwait(false);
+            if (batch?.Count > 0)
```
zarusz, line 238: I think you still could detect if this is the last message on line 220 and if so, skip the continue, which would allow us to eliminate these if altogether and lines 240-241. Same applies to AEH implementation.

src/SlimMessageBus.Host.Outbox/Repositories/IOutboxRepository.cs
```diff
@@ -3,8 +3,9 @@
 public interface IOutboxRepository
 {
     Task Save(OutboxMessage message, CancellationToken token);
-    Task<int> TryToLock(string instanceId, DateTime expiresOn, CancellationToken token);
-    Task<IReadOnlyList<OutboxMessage>> FindNextToSend(string instanceId, CancellationToken token);
+    Task<IReadOnlyCollection<OutboxMessage>> LockAndSelectAsync(string instanceId, int batchSize, bool tableLock, TimeSpan lockDuration, CancellationToken token);
```
zarusz, line 6: Is the Async suffix in the name intended?

src/SlimMessageBus.Host.Outbox/IOutboxLockRenewalTimer.cs
```diff
@@ -0,0 +1,12 @@
+﻿namespace SlimMessageBus.Host.Outbox;
+
+public interface IOutboxLockRenewalTimer : IDisposable
+{
+    bool Active { get; }
+    public string InstanceId { get; }
+    public TimeSpan LocKDuration { get; }
```
zarusz, line 7: `LockDuration`

src/SlimMessageBus.Host.Outbox/Interceptors/OutboxSendingTask.cs
```diff
@@ -179,99 +182,122 @@ private async Task Run()
...
-            }
-        } while (!idleRun && !ct.IsCancellationRequested);
+
+            lockRenewalTimer.Stop();
+        } while (!cts.Token.IsCancellationRequested && runAgain);
 
         return count;
     }
 
+    internal async Task<(bool Success, int Published)> DispatchBatchAsync(IOutboxRepository outboxRepository, IMessageBusBulkProducer producer, IMessageBusTarget messageBusTarget, IReadOnlyCollection<EnvelopeWithId> batch, string busName, string path, CancellationToken cancellationToken)
```
zarusz, line 274: Is the Async suffix intended here?

src/SlimMessageBus.Host.Outbox/Interceptors/OutboxSendingTask.cs
```diff
@@ -179,99 +182,122 @@ private async Task Run()
...
+
+        // TOOD: Enclose in a transaction
+        var results = await producer.ProduceToTransport(batch, path, messageBusTarget, cancellationToken).ConfigureAwait(false);
+        if (cancellationToken.IsCancellationRequested && results.Dispatched.Count == 0)
+        {
+            // if cancellation has been requested, only return if no messages were published
+            return (false, 0);
+        }
+
+        var updatedIds = results.Dispatched.Select(x => x.Id).ToList();
```
zarusz, line 286: Perhaps making this HashSet would be slightly better?

src/SlimMessageBus.Host.Outbox/Interceptors/OutboxSendingTask.cs
```diff
@@ -179,99 +182,122 @@ private async Task Run()
...
+        await outboxRepository.UpdateToSent(updatedIds, CancellationToken.None).ConfigureAwait(false);
```
zarusz, line 287: CancellationToken.None is on purpouse so that we exit before the messages are marked as sent?

src/SlimMessageBus.Host.Outbox/OutboxLockRenewalTimer.cs
```diff
@@ -0,0 +1,142 @@
...
+        }
+
+        try
+        {
+
+            try
+            {
+                try
+                {
+                    if (!await _outboxRepository.RenewLock(InstanceId, LocKDuration, _cancellationToken))
```
zarusz, line 87: I wonder about concurrency here. The OutboxRepository uses 1 SQL connection. This timer might execute at the same time as the message send loop and cause a race condition on the SQL connection (will be hard to find and debug if this happens).

Perhaps  we'd need to synchronize the repository instance OR better to set some flag (send signal) for the main outbox task loop to call the renew lock from within the loop?

src/Tests/SlimMessageBus.Host.Test/MessageBusBaseTests.cs
zarusz: Could you Skip the test rather than remove it? 
I do want to rewrite it and still see value in it.

src/SlimMessageBus.Host.Outbox/Repositories/IOutboxRepository.cs
EtherZa: hmmm... I never noticed that SMB was not following the convention. Updated.

src/SlimMessageBus.Host.Outbox/Interceptors/OutboxSendingTask.cs
EtherZa: Changed.

src/SlimMessageBus.Host.Outbox/Interceptors/OutboxSendingTask.cs
EtherZa: Correct. No transaction support yet, so must attempt to keep the db in sync with what has been dispatched already.

zarusz: If this is the place you referred to when mentioning Distributed Transaction Coordinator (DTC) then here are my thoughts:
1. I don't want to engage DTC or similar mechanism here to over complicate things and coordinate the transport system and SQL db
2. If anything the message will be sent again, and so will result with at least once guarantee here which should be fine in the rather unlikely event that after a successful publish to transport the update messages to sent in SQL fails.

If however, you're thinking about spaning the select next messages from SQL and marking the messages as sent in SQL  with a db transaction then it makes sense to do, but still not sure if there is danger here not doing so.

src/SlimMessageBus.Host.Outbox/OutboxLockRenewalTimer.cs
EtherZa: There are three instances of `OutboxRepository` in play through a bit of scope juggling.
1. One per message pipeline (either root or scoped dependent on configuration)
2. `OutboxSendTask` is a singleton that starts a new scope in `Run()` which is then used to resolve `OutboxRepository` for locking/processing/updating messages.
3. `OutboxLockRenewalTimerFactory` is a also singleton that creates a new scope on construction that is used to host `OutboxLockRenewalTimer`.

```
Incoming message
	=> new scope (optional)
		=> Interceptor pipeline
			=> SqlOutboxRepository.Save

OutboxSendTask (Singleton)
	=> new scope
		=> SendMessages()
			=> SqlOutboxRepository.LockAndSelect
			=> SqlOutboxRepository.IncrementDeliveryAttempt
			=> DispatchBatchAsync()
				=> SqlOutboxRepository.UpdateToSent
				=> SqlOutboxRepository.IncrementDeliveryAttempt

OutboxLockRenewalTimerFactory (Singleton)
	=> new scope
		=> OutboxLockRenewalTimer (scope)
			=> SqlOutboxRepository.RenewLock
```

It would be better to use `IDbContextFactory<>` to retrieve an instance from the pool. Unfortunately, default registrations are incompatible with each other. A future PR should address this TBH. Developer focus is likely to shift to the pooled factory implementation for performance purposes.

EtherZa:
@zarusz  Comments resolved, test added and rebased

src/SlimMessageBus.Host.Outbox/OutboxLockRenewalTimer.cs
zarusz: That makes sense. I thought OutboxSendTask uses the same instance as OutboxLockRenewalTimerFactory.

Also great idea with `IDbContextFactory<>` in the future.

zarusz:
Thanks for this massive PR @EtherZa!

EtherZa:
> Thanks for this massive PR @EtherZa!
 
@zarusz It's a pleasure. Thanks for the help and library :)

PR #262, opened 2024-06-13
[Integration testing] Containers for local development

EtherZa:
Testing if you aren't @zarusz can be a little painful with the test configurations tailored for private resources ;)

Please consider this PR. It allows for a full test coverage running everything locally except for Azure Service Bus and Azure Event Hub which don't have containers available.

To ease some of the integration, Kafka and Mqtt are both hosted without certs. It does require some minor changes to some of the bootstrapping for the integration tests, but won't have any impact (beyond configuration) for the build pipeline.

I understand that build pipeline will be broken with the change, but that should be easily resolved by updating the devops version of secrets.txt with the sample provided.

## Local development environment using docker
1. Copy `src/secrets.txt.sample` to `src/secrets.txt`
2. Add connection strings for personal instances of the below to secrets.txt
	- Azure Service Bus
	- Azure Event Hub
3. there is no auto-provisioning available for event hubs, so the below event hubs and consumer groups will need to be created manually (event hub => consumer group)
- test-ping => subscriber
- test-echo => handler
- test-echo-resp => response-reader
4. open a terminal to src/Infrastructure
5. execute `docker compose up --force-recreate -V` or `./compose.sh`
6. Run tests and smile

### DIFF
--- a/src/Infrastructure/README.md
+++ b/src/Infrastructure/README.md
@@ -1,0 +1,33 @@
+# Local infrastructure
+ 
+Local containers to facilitate testing for all messages buses (except for Azure Event Hub and Azure Service Bus) can be created and hosted using [Docker Desktop](https://www.docker.com/products/docker-desktop/). 
+
+No volumes have been configured so as to provide clean instances on each run.
+ 
+ ```
+src/Infrastructure> docker compose up --force-recreate -V
+ ```
+ or
+ ```
+ > ./infrstructure.sh
+ ```
+
+## Configuration
+
+Personal instances of Azure resources are required as containers are not available for those services.
+
+1. Copy `src/secrets.txt.sample` to `src/secrets.txt`
+2. Supply personal instances of Azure Event Hub and Azure Service Bus
+3. Add connection strings to `src/secrets.txt` for 
+	- Azure Service Bus
+	- Azure Event Hub
+
+
+### Azure Event Hub
+The transport/plug-in does not provide topology provisioning. The below event hubs and consumer groups will need to be manually added to the Azure Event Hub instance for the integration tests to run.
+
+| Event Hub			| Consumer Group	|
+|---------------------------------------|
+| test-ping			| subscriber		|
+| test-echo			| handler			|
+| test-echo-resp	| response-reader	|
--- a/README.md
+++ b/README.md
@@ -354,7 +354,7 @@ NuGet packages end up in `dist` folder
 
 ## Testing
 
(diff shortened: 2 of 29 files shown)


zarusz:
I agree that having test containers where possible is a step forward. 

Let me think about this some more and perhaps accept this PR into a side master branch and then iterate on combining with the private cloud resources that are currently used for testing.

Again, thanks for your help and effort.

EtherZa:
> Again, thanks for your help and effort.

@zarusz It's a pleasure. We all win here

>Let me think about this some more and perhaps accept this PR into a side master branch and then iterate on combining with the private cloud resources that are currently used for testing.

The first approach I took was to just add support for environment variables to `BaseIntegrationTest.Configuratoin`. Combining that with a .runsettings worked well until I tried to get it to work with live unit testing. Visual Studio ignores .runsettings in LUT.

This approach is still not perfect as `secrets.txt` is not included with the LUT shadow copy. Manually copying the file to the LUT directory gets the tests running successfully at least.

src/Infrastructure/README.md
```diff
@@ -0,0 +1,29 @@
+# Infrastructure for local development
+ 
+ Local containers to facilitate testing for all messages buses (except for Azure Event Hub and Azure Service Bus).
```
zarusz, line 3: Space in front of paragraph

src/Infrastructure/README.md
```diff
@@ -0,0 +1,29 @@
+# Infrastructure for local development
```
zarusz, line 1: On the main page README.md there is some notes about running tests locally.
Either make a note to link form that to this README, else add some section inside `intro.t.md`.

zarusz:
@EtherZa I've adjusted the build.yaml and added the new secrets in GitHub to have both local and cloud tests running.
Please rebase against master. 

Hopefully, it will work after one shot :)

src/Infrastructure/README.md
EtherZa: Updated the `Testing` section in /README.md and linked to the infrastructure README.md

I have also moved/renamed the compose bash script to the repo root as Infrastructure.sh which will make it easier to initiate

```
/> ./Infrastructure.sh
```

EtherZa:
Works like a charm - local tests are all green. 

The Kafka and Event Hub tests intermittently need an extra run as there appears to be a race condition between each set; but that is a minor issue for now.

I am also hoping that you will forgive the code coverage complaint that SonarCloud is raising - it is the empty/commented line predicate in the secrets file reader. The build pipeline clearly makes use of neither condition :)

Thank you @zarusz. This will make a big difference.

PR #263, opened 2024-06-14
[Host.Outbox] Ensure SQL tables are provisioned and baseline test

zarusz:
- Ensure SQL tables are provisioned before the first publish (and without having to start consumers).
- Send Created bus lifecycle event to be able to hook into Master bus creation
- Add baseline performance test for Outbox.

The test (`OutboxBenchmarkTests`) would be run locally against the same machine (GitHub Actions agent). It collects the times it took to produce and then consumes N messages. We can use the times to compare if the new enhancements are better.  In the future more types of use cases could be added to baseline the Outbox from different angles.

For example, a run from my local machine to the Azure Service Bus + Azure SQL instance, for 1000 messages:

With Outbox:
```
[12:59:34 INF] SlimMessageBus.Host.Outbox.DbContext.Test.OutboxBenchmarkTests Message Publish took: "00:00:18.2214666"
[12:59:34 INF] SlimMessageBus.Host.Outbox.DbContext.Test.OutboxBenchmarkTests Message Consume took: "00:01:42.5709066"
```

Without Outbox:
```
[12:57:33 INF] SlimMessageBus.Host.Outbox.DbContext.Test.OutboxBenchmarkTests Message Publish took: "00:00:02.2195892"
[12:57:33 INF] SlimMessageBus.Host.Outbox.DbContext.Test.OutboxBenchmarkTests Message Consume took: "00:00:06.5525862"
```

### DIFF
--- a/infrastructure.ps1
+++ b/infrastructure.ps1
@@ -1,0 +1,1 @@
+docker compose -f src/Infrastructure/docker-compose.yml up --force-recreate -V
--- a/src/Host.Plugin.Properties.xml
+++ b/src/Host.Plugin.Properties.xml
@@ -4,7 +4,7 @@
 	<Import Project="Common.NuGet.Properties.xml" />
 
 	<PropertyGroup>
-    <Version>2.3.5</Version>
+    <Version>2.4.0-rc1</Version>
   </PropertyGroup>
 
 </Project>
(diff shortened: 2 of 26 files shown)


zarusz:
@EtherZa I have fine-tuned the ASB settings and made the test run with outbox vs without outbox to understand the relative overhead. See the new measurements above.

PR #265, opened 2024-06-16
#263 outbox benchmark tweaks

EtherZa:
@zarusz I picked up a few issues when running the benchmarks, so threw together a couple fixes

* `OutboxSendingTask.OnBusLifecycle()` can return a null value - exception thrown in `MessageBusBase.OnBusLifecycle()`
* `OutboxBenchmarkTests.SetupServices()` - `mbb.AutoStartConsumersEnabled(false);` only propagates if is set before adding a child bus. Child bus was immediately active causing some messages to be consumed before the consumption timer was started.
* `BaseIntegrationTest.EnsureConsumersStarted()` just watched for a consumer to start, but didn't start one. 
* Refactored `OutboxSendingTask` so that `SendMessage()` can be called directly (`internal` modifier). Publishing time can now be isolated from consumption time. Same logic, but shifted the primary loop from `Run()` to `SendMessages()`.
* Drop the db and recreate it instead of a clean up - less "error" prone while debugging migration scripts (may change with every test run)
* Outbox migration is also manually run after rebuilding db as I picked up that the test publishes messages before the migration has finished running

I have picked up a few logical errors in #260 which is the reason for the delay.


src/Tests/SlimMessageBus.Host.Outbox.DbContext.Test/OutboxBenchmarkTests.cs
```diff
@@ -118,6 +130,24 @@ await using (unitOfWorkScope as IAsyncDisposable)
...
+        // publish messages
+        var outboxPublishTimerElapsed = TimeSpan.Zero;
+        if (_useOutbox)
+        {
+            var outboxSendingTask = ServiceProvider.GetRequiredService<OutboxSendingTask>();
+            var outputRepository = ServiceProvider.GetRequiredService<IOutboxRepository>();
+
+            var outboxTimer = Stopwatch.StartNew();
+            var publishCount = await outboxSendingTask.SendMessages(ServiceProvider, outputRepository, CancellationToken.None).ConfigureAwait(false);
+            outboxTimer.Stop();
```
zarusz, line 142: I don't think you need to stop it? The next line will already read the Elapsed. One line less to worry :)

src/Tests/SlimMessageBus.Host.Outbox.DbContext.Test/OutboxBenchmarkTests.cs
```diff
@@ -126,12 +156,12 @@ await using (unitOfWorkScope as IAsyncDisposable)
         await store.WaitUntilArriving(newMessagesTimeout: 5, expectedCount: events.Count);
 
         // assert
-
         var consumeTimerElapsed = consumptionTimer.Elapsed;
 
         // Log the measured times
-        Logger.LogInformation("Message Publish took: {Elapsed}", publishTimerElapsed);
-        Logger.LogInformation("Message Consume took: {Elapsed}", consumeTimerElapsed);
+        Logger.LogInformation("Message Publish took       : {Elapsed}", publishTimerElapsed);
```
zarusz, line 162: Just a note. I was thinking that in the future we could also collect the measurements into a txt file. The file could then be part of the build output artifact - so quickly compare various implementations (and assess if its better).

src/SlimMessageBus.Host.Outbox/SlimMessageBus.Host.Outbox.csproj
```diff
@@ -1,20 +1,26 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
-  <Import Project="../Host.Plugin.Properties.xml" />
+	<Import Project="../Host.Plugin.Properties.xml" />
```
zarusz, line 3: Just a note. In the future might be good to tweak the `.editorconfig` to standardize tab width.

zarusz:
@EtherZa  looks good to me. Let me know once you've done with the changes and resolved the comments.

I've re-run the tests, there seems to be some transient connection to Azure SQL.

zarusz:
I believe is a clash between the two Outbox tests and PRs that apply different sets of migrations. Perhaps the delete from the migrations table does not serve well.

I dropped the two tables from the Azure SQL instance to let them be recreated and migration table re-populated.

```sql
DROP TABLE [dbo].[__EFMigrationsHistory]
DROP TABLE [dbo].[IntTest_Customer]
```

The last error from the integration tests:

```
[xUnit.net 00:01:38.65]         [17:50:44 ERR] Microsoft.EntityFrameworkCore.Database.Command Failed executing DbCommand (161ms) [Parameters=[], CommandType='"Text"', CommandTimeout='30']
[xUnit.net 00:01:38.65]         ALTER TABLE [IntTest_Customer] ADD [UniqueId] nvarchar(max) NULL;
  Failed SlimMessageBus.Host.Outbox.DbContext.Test.OutboxTests.Given_CommandHandlerInTransaction_When_ExceptionThrownDuringHandlingRaisedAtTheEnd_Then_TransactionIsRolledBack_And_NoDataSaved_And_NoEventRaised(transactionType: SqlTransaction, busType: AzureSB, messageCount: 100) [17 s]
  Error Message:
   Microsoft.Data.SqlClient.SqlException : Column names in each table must be unique. Column name 'UniqueId' in table 'IntTest_Customer' is specified more than once.
```

src/Tests/SlimMessageBus.Host.Outbox.DbContext.Test/OutboxBenchmarkTests.cs
```diff
@@ -76,14 +86,16 @@ public async Task Given_EventPublisherAndConsumerUsingOutbox_When_BurstOfEventsI
         // arrange
         _useOutbox = useOutbox;
 
-        await PerformDbOperation(async context =>
+        await PerformDbOperation(async (context, outboxMigrationService) =>
         {
             // migrate db
+            await context.Database.EnsureDeletedAsync();
```
zarusz, line 92: This is causing issues with the failing int test.
This is an Azure SQL database and should not be deleted like that.

Could you not move the EraseTableIfExists above the MigrateAsync, or even crop the outbox and migrations table instead?

```
[xUnit.net 00:01:05.76]     SlimMessageBus.Host.Outbox.DbContext.Test.OutboxBenchmarkTests.Given_EventPublisherAndConsumerUsingOutbox_When_BurstOfEventsIsSent_Then_EventsAreConsumedProperly(useOutbox: False, messageCount: 1000) [FAIL]
[xUnit.net 00:01:05.77]       Microsoft.Data.SqlClient.SqlException : Database 'SlimMessageBus_Outbox' already exists. Choose a different database name.
[xUnit.net 00:01:05.77]       Stack Trace:
```

src/SlimMessageBus.Host.Outbox/SlimMessageBus.Host.Outbox.csproj
EtherZa: I have added the .csproj default (double space) to the .editorconfig

src/Tests/SlimMessageBus.Host.Outbox.DbContext.Test/OutboxBenchmarkTests.cs
EtherZa: I have replaced it with a script to delete all user defined objects in the db instead.

EtherZa:
@zarusz One other enhancement that you may want to consider is to use a unique topic with `AutoDeleteOnIdle` for each test execution to ensure that no messages remain the subscription from a previous failed test run.

ie. something like
```

            mbb.AddChildBus("ExternalBus", mbb =>
            {
                var topic = $"tests.outbox-benchmark/customer-events/{Guid.NewGuid():N}";
                mbb
                    .WithProviderServiceBus(cfg =>
                    {
                        cfg.ConnectionString = Secrets.Service.PopulateSecrets(configuration["Azure:ServiceBus"]);
                        cfg.PrefetchCount = 100; // fetch 100 messages at a time
                        cfg.TopologyProvisioning.CreateTopicOptions = opt => opt.AutoDeleteOnIdle = TimeSpan.FromMinutes(5);
                    })
```

I did not make the change now as it will require manually deleting the existing topic before it can be used as a root for this pattern.

src/Tests/SlimMessageBus.Host.Outbox.DbContext.Test/DatabaseFacadeExtensions.cs
```diff
@@ -0,0 +1,217 @@
+﻿namespace SlimMessageBus.Host.Outbox.DbContext.Test;
+
+using Microsoft.Data.SqlClient;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Infrastructure;
+
+public static class DatabaseFacadeExtensions
+{
+    public static async Task EnsureDatabaseIsEmptyAsync(this DatabaseFacade database, string schema = "dbo", CancellationToken cancellationToken = default)
```
zarusz, line 9: I'd rather drop whole tables *selectively* before the test rather than scan the whole DB and remove all objects. This might touch some Azure SQL objects but my main point is that it worsens the test isolation (we might bring back concurrent test loads at some point or for selected areas).

zarusz: Also, ping me via email (on my GitHub profile page) - want to share something with you to help with debugging.

EtherZa: > I'd rather drop whole tables selectively before the test rather than scan the whole DB and remove all objects. This might touch some Azure SQL objects but my main point is that it worsens the test isolation (we might bring back concurrent test loads at some point or for selected areas).

Yeah, it took a little longer than I would have liked to admit to realise that the db was shared and tests were using it concurrently. I pivoted to schema deployments instead - one schema per class that can be dropped/recreated on a test run. It would allow for full spin up without the risk of doing something unintended.

I would require `MigrateAsync()` to be dropped in favour of `EnsureCreatedAsync()` - no `__EFMigrationsHistory` table and no migrations in integration tests but a clean, isolated setup per execution. It would be opt in as well, so any ordered tests could still do what they would need to do.

How do you feel about this approach?

zarusz: You can give it a try.

I've tried to isolate the SMB outbox table and outbox migration table by simply assigning a name tied with the OutboxBenchmarkTests (added a prefix). The same could be done with EF migrations table too:
https://learn.microsoft.com/en-us/ef/core/managing-schemas/migrations/history-table
I did not do it at the time but likely that could have done the trick and given us the sufficient isolation we wanted.

EtherZa: Okay. So that was an exercise in pure frustration :/

The short answer is that the `Outbox` models can live safely in different schemas per test and the schemas can be easily refreshed/isolated per test run. EF Core however, is a different story. 

EF simply does not support [multi-tenancy by schema](https://learn.microsoft.com/en-us/ef/core/miscellaneous/multitenancy#multiple-schemas). Further to this, modifying the table name/schema is problematic as the last value will be retained for all future model interactions (updated singleton). This limitation leaves safe concurrency of tests using the same model and database instance unfeasible.

EtherZa:
This is the best I can come up with for a single db implementation.

Use the same schema for both EF and outbox migrations; dropping it on every test. Any test that uses the same db context/schema will need to run in series unfortunately.

On the upside, it will create a completely clean copy of the db for every test run, as well as clean instance of the service bus (with a 5 min auto delete). Repeatable and isolated from other tests while using a single SQL database instance.

If you would like to go the full concurrency/parallel execution route, it might be worth looking at [test containers](https://testcontainers.com/). ASB/AEH can use different topics to achieve the same thing without containers being available.

src/Tests/SlimMessageBus.Host.Outbox.DbContext.Test/DatabaseFacadeExtensions.cs
```diff
@@ -0,0 +1,252 @@
...
+            await database.EnsureSchemaIsEmptyAsync(schema, cancellationToken);
+            await database.ExecuteSqlRawAsync(sql, [schema], cancellationToken);
+        }
+        catch (SqlException ex) when (ex.Number == 4060)
+        {
+            // db does not exist
+        }
+    }
+
+    public static async Task EnsureSchemaIsEmptyAsync(this DatabaseFacade database, string schema, CancellationToken cancellationToken = default)
```
zarusz, line 42: This method seems unused now, if so please remove it.

zarusz:
@EtherZa Well done with the changes - it looks better & clean!

Issue #275, opened 2024-06-28
[Host.Outbox] Polling causes an unnessary delay / notify OutboxSendingTask that the outbox has been populated


EtherZa:
When a message is sent to the outbox for deferred publishing, the `OutboxSendingTask` service can be notified that a new message(s) is waiting to be processed so as to avoid waiting for the timer to elapse.

zarusz:
Closed via #276 
See the 2.4.0-rc10 release (or newer).

Issue #110, opened 2022-07-09
[Host.Kafka] Ability to provision topology (topic/consumer groups) for Kafka


zarusz:
It would be good if SlimMessageBus Kafka transport is able to auto-provision the necessary topic and consumer groups.

Example for the underlying driver:
https://github.com/confluentinc/confluent-kafka-dotnet/blob/master/examples/AdminClient/Program.cs

PR #276, opened 2024-06-28
#275 Notify outbox service that a message is available for publishing

EtherZa:
On sending a message to the outbox, the `OutboxForwardingPublishInterceptor` will notify the `OutboxSendingTask` that a publishing loop should begin. If the `OutboxSendingTask` is already publishing messages, the notification will be ignored.

The notification is only accepted if the service is currently active. 

The original `PollIdleSleep` configuration is still used should no notification be received. The sleep timer is both a fail safe for the process and will initiate publishing in a distributed environment (no cross app communication available).

**Note:** notifications are sent from `OutboxForwardingPublishInterceptor` on disposal so as to ensure they are raised only once any transactions have been completed.


EtherZa:
I've updated the ASB tests to provision unique topic/queue names per execution (with a 5 min auto-delete on idle). I think that just leaves the Kafka tests as being a little flakey now.

docs/plugin_outbox.md
```diff
@@ -203,12 +203,12 @@ When applied on the (child) bus level then all consumers (or handlers) will inhe
...
+- On the interceptor being disposed (at then end of a message lifecycle), the outbox service will be notified that a message is waiting to be published if at least one messages was placed in the outbox during the message processing lifetime. Alternatively, on expiry of the `PollIdleSleep` period following outbox processing, the same process will be initiated.
 
-- Message added to the `Outbox` table are initially owned by the respective service instance that created it, the message has a lock that expires at some point in time (driven by settings). Every service instance task attempts to publish their owned messages which happens in order of creation (this ensures order of delivery within the same process).
+- If the configuration setting `ManitainSequence` is set to `true`, only one application instance will be able to lock messages for delivery. This ensure messages are delivered in the original order to the service bus at the expense of delivery throughput. If `false`, each distributed instance will place an exclusive lock on `PollBatchSize` messages for concurrent distribution. This will greatly increase throughput at the expense of the FIFO sequence.
```
zarusz, line 208: Typo: Maintain

src/SlimMessageBus.Host.Outbox/Configuration/OutboxSettings.cs
```diff
@@ -15,7 +15,7 @@ public class OutboxSettings
     /// <summary>
     /// Sleep time of the outbox polling loop if there were no messages to process in previous database poll.
     /// </summary>
-    public TimeSpan PollIdleSleep { get; set; } = TimeSpan.FromSeconds(1);
+    public TimeSpan PollIdleSleep { get; set; } = TimeSpan.FromMinutes(1);
```
zarusz, line 18: 1 minute might be too long but then with the new mechanism to notify that you're introducing it should be good and acting as a failsafe.

src/SlimMessageBus.Host.Outbox/AsyncManualResetEvent.cs
```diff
@@ -0,0 +1,59 @@
+﻿namespace SlimMessageBus.Host.Outbox;
+
+public sealed class AsyncManualResetEvent
```
zarusz, line 3: I suppose there isn't anything in the framework that would give us such a primitive (unless we used a 3rd party lib).

I wonder about the potential re-allocation of `TaskCompletionSource` with every outbox loop run. While it should not be a heavy object to alloc, thinking about it perhaps we could use SemaphoreSlim which also has an WaitAsync supporting timouts. Could you contemplate on that? If this were to work it could help us avoid memory pressure with every loop run.

src/SlimMessageBus.Host.Outbox/Configuration/MessageBusBuilderExtensions.cs
```diff
@@ -34,6 +36,7 @@ public static MessageBusBuilder AddOutbox(this MessageBusBuilder mbb, Action<Out
 
             services.AddSingleton<OutboxSendingTask>();
             services.TryAddEnumerable(ServiceDescriptor.Singleton<IMessageBusLifecycleInterceptor, OutboxSendingTask>(sp => sp.GetRequiredService<OutboxSendingTask>()));
+            services.TryAddEnumerable(ServiceDescriptor.Singleton<IOutboxNotificationService, OutboxSendingTask>(sp => sp.GetRequiredService<OutboxSendingTask>()));
```
zarusz, line 39: On one hand in the OutboxNotificationService a 1 instance of IOutboxNotifiationService is expected, on the other hand we register it as if multiple service implementations could be registered here.

(use `TryAdd()` instead)

zarusz:
@EtherZa  good feature and set of changes!

I've left some minor feedback. Once you address and are ready then I am happy to merge.

On the Kafka topic creation, we would have to implement this feature #110 to make this work.

src/SlimMessageBus.Host.Outbox/AsyncManualResetEvent.cs
EtherZa: There isn't anything in the framework that I am aware of that does the same thing. It is similar in implementation to that of [Nito.AsyncEx](https://github.com/StephenCleary/AsyncEx/blob/0361015459938f2eb8f3c1ad1021d19ee01c93a4/src/Nito.AsyncEx.Coordination/AsyncManualResetEvent.cs).

I ran a benchmark against a `SemaphoreSlim` variant with the below results.

```
| Method        | Mean     | Error    | StdDev   | Allocated |
|-------------- |---------:|---------:|---------:|----------:|
| TestTcs       | 15.54 ms | 0.027 ms | 0.025 ms |     683 B |
| TestSemaphore | 15.54 ms | 0.048 ms | 0.042 ms |    1139 B |
```

src/SlimMessageBus.Host.Outbox/Configuration/OutboxSettings.cs
EtherZa: The  timed execution shouldn't ever be necessary on a single instance app. The impact is going to be more on distributed applications when the `MaintainSequence` setting is false (concurrent processing) or if there are insertions into the outbox via the backdoor.

How long do you think the default period should be?

src/SlimMessageBus.Host.Outbox/AsyncManualResetEvent.cs
zarusz: Did the test include allocations of TCS per each operation? Because this is my main concern.

src/SlimMessageBus.Host.Outbox/Configuration/OutboxSettings.cs
zarusz: I think it's fine as you made it 1h. I was just thinking out loud ;)

EtherZa:
@zarusz Thanks for the feedback. I've fixed the typo (and made the change in the _t.md). The current `AsyncManualResetEvent` implementation looks to have a lower impact than a `SemaphoreSlim` variant (benchmark on comment).

Please let me know if you would like the default idle timeout to change and/or if you come across anything else. Otherwise, I think that it is good to go from my perspective.

src/SlimMessageBus.Host.Outbox/AsyncManualResetEvent.cs
EtherZa: Here's the implementation/benchmark - https://gist.github.com/EtherZa/dd4295ee5de086683a4548b386add0cf

Let me know if you would like anything changed.

zarusz: Wow, thanks for the details. Interesting that Semaphore Wait alloc more than TCS alloc.

EtherZa: I was a little surprised myself tbh.

Issue #251, opened 2024-04-19
[Host | Feature request] Toggle consumers by health check tag(s)


EtherZa:
_TLDR; associate consumers with health checks to enable/disable them as the status of dependent services change_

I haven't seen this implemented in a competing library, and it may be beyond the "slim" scope, but all the same it may be a feature that garners interest. It would certainly be useful in my environment.

Consumers often require different downstream services to fulfil their execution. One may need a specific database instance, another network storage, while a third may require a third party API (or a combination of services) to be available. When one of these dependencies become unavailable, the consumer needs to be resilient and handle the failure gracefully or fail successively before the message ends up in the dead letter queue.

If consumers were to be associated with the health checks of their dependent services, then they could be individually started/stopped on a health status change. Disabling only the impacted consumer(s) would ease the impact while allowing uneffected consumers to continue processing messages unimpeded.

A potential solution is to have a proxy sitting between `MessageBusBase` and the `Consumers` to intercept `Start`/`Stop` events while monitoring health checks (tags) to start/stop the consumer as appropriate.

zarusz:
@EtherZa Sounds like a good use case. If you willing to contribute I am happy to review a PR. Perhaps it's a circuit breaker even.

EtherZa:
@zarusz I'll take a look at it, but it may take a little longer to implement.

PR #282, opened 2024-07-17
#251 Health check circuit breaker

EtherZa:
Addition of a health check circuit breaker for consumers (`SlimMessageBus.Host.CircuitBreaker.HealthCheck`). 

Consumers can be optionally associated with one or more tags (and severities) in order to temporarily pause the consumer while a health check is failing. 

The `HealthCheckBackgroundService` taps into the published health reports to distribute health status changes generated by the `Microsoft.Extensions.Diagnostics.HealthChecks` library. It is added as a hosted service to catch reports outside of the life time of the `SlimMessageBus`.

```cs
    // add health checks with tags
    builder.Services
        .AddHealthChecks()
        .AddCheck<StorageHealthCheck>("Storage", tags: ["Storage"]);
        .AddCheck<SqlServerHealthCheck>("SqlServer", tags: ["Sql"]);

    builder.Services
        .AddSlimMessageBus(mbb => {
            ...

            mbb.Consume<Message>(cfg => {
                ...

                // configure consumer to monitor tag/state
                cfg.PauseOnUnhealthyCheck("Storage");
                cfg.PauseOnDegradedHealthCheck("Sql");
            })
        })
```
---

To add the functionality, a few questionable changes have been made. Please shout if you are not comfortable with any of them/would prefer an alternate implementation.

* `IEnumerable<AbstractConsumerSettings>` is required by `AbstractConsumer's` constructor (retrieve associated circuit breaker configuration)
* `AbstractConsumerSettings` maintains a reference to the associated `MessageBusSettings` (`IServiceProvider` to create the `IConsumerCircuitBreaker` instances and to retrieve the associated bus name for logging)
* `AbstractConsumerBuilder` extends `IHasPostConfigurationActions` that will be executed by the `MessageBusBuilder` pipeline (circuit breakers service registrations)

A sample application has been included (`Sample.CircuitBreaker.HealthCheck`) to demonstrate the functionality with health checks that fail randomly.

### Ideas for future improvements
Although I have no plans on doing this now, the functionality could be extended by incorporating an `IConsumerInterceptor` to watch for exceptions and, on catching one, initiate an ad hoc health check (once per exception type/period/etc). The evaluation period could then be reduced from the standard health check publishing interval to a reactive check based on an application failure backed up by the default cadence.


EtherZa:
@zarusz It looks like sql may be experiencing an issue.

```
[07:36:13 ERR] SlimMessageBus.Host.Outbox.Services.OutboxSendingTask Error while processing outbox messages
 Microsoft.Data.SqlClient.SqlException (0x80131904): A severe error occurred on the current command.  The results, if any, should be discarded.
 Operation cancelled by user.
    at Microsoft.Data.SqlClient.SqlConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
    at Microsoft.Data.SqlClient.TdsParser.ThrowExceptionAndWarning(TdsParserStateObject stateObj, SqlCommand command, Boolean callerHasConnectionLock, Boolean asyncClose)
    at Microsoft.Data.SqlClient.TdsParser.TryRun(RunBehavior runBehavior, SqlCommand cmdHandler, SqlDataReader dataStream, BulkCopySimpleResultSet bulkCopyHandler, TdsParserStateObject stateObj, Boolean& dataReady)
    at Microsoft.Data.SqlClient.SqlDataReader.TryConsumeMetaData()
    at Microsoft.Data.SqlClient.SqlDataReader.get_MetaData()
    at Microsoft.Data.SqlClient.SqlCommand.FinishExecuteReader(SqlDataReader ds, RunBehavior runBehavior, String resetOptionsString, Boolean isInternal, Boolean forDescribeParameterEncryption, Boolean shouldCacheForAlwaysEncrypted)
    at Microsoft.Data.SqlClient.SqlCommand.CompleteAsyncExecuteReader(Boolean isInternal, Boolean forDescribeParameterEncryption)
    at Microsoft.Data.SqlClient.SqlCommand.EndExecuteReaderInternal(IAsyncResult asyncResult)
    at Microsoft.Data.SqlClient.SqlCommand.EndExecuteReaderAsync(IAsyncResult asyncResult)
    at System.Threading.Tasks.TaskFactory`1.FromAsyncCoreLogic(IAsyncResult iar, Func`2 endFunction, Action`1 endAction, Task`1 promise, Boolean requiresSynchronization)
 --- End of stack trace from previous location ---
    at SlimMessageBus.Host.Outbox.Sql.SqlOutboxRepository.LockAndSelect(String instanceId, Int32 batchSize, Boolean tableLock, TimeSpan lockDuration, CancellationToken token) in /_/src/SlimMessageBus.Host.Outbox.Sql/SqlOutboxRepository.cs:line 54
    at SlimMessageBus.Host.Outbox.Services.OutboxSendingTask.SendMessages(IServiceProvider serviceProvider, IOutboxRepository outboxRepository, CancellationToken cancellationToken) in /_/src/SlimMessageBus.Host.Outbox/Services/OutboxSendingTask.cs:line 245
    at SlimMessageBus.Host.Outbox.Services.OutboxSendingTask.Run() in /_/src/SlimMessageBus.Host.Outbox/Services/OutboxSendingTask.cs:line 173
 ClientConnectionId:39c33641-24b9-4302-92bb-4b475a76b5ad
 ```

File: build/tasks.ps1 [hunk 1]
```diff
@@ -41,6 +41,8 @@ $projects = @(
 	"SlimMessageBus.Host.Outbox.Sql",
 	"SlimMessageBus.Host.Outbox.DbContext",
 
+	"SlimMessageBus.Host.CircuitBreaker.HealthCheck",
```
zarusz, line 44: Padding - tabs or spaces

docs/intro.md
```diff
@@ -258,6 +259,32 @@ await consumerControl.Stop();
```
zarusz, line 259: I believe we should slate these new features for the 3.x release. It's been waiting on a separate branch for few weeks and we need to release it soon.

zarusz:
@EtherZa excellent feature and nice proposed API. Agreed that in the future a consumer interceptor could also observe for failure rates and proactively switch off the circuit - that could be added later.

I will be able to give it some review in 1.5 weeks as I am out now without my laptop.

build/tasks.ps1
EtherZa: Tab to match

tabs

EtherZa:
@zarusz No worries. I just hope that's a pleasure trip :)

docs/intro.t.md
```diff
@@ -258,6 +259,32 @@ await consumerControl.Stop();
 
 > Since version 1.15.5
 
+#### Health check circuit breaker
+
+Consumers can be associated with health check tags to enable or disable the consumer as the check's status changes.
```
zarusz, line 264: Maybe elaborate one more sentence that the health checks are based on on Microsoft Diagnostics HealthChecks

src/SlimMessageBus.Host.Configuration/Builders/MessageBusBuilder.cs
```diff
@@ -26,6 +26,8 @@ public class MessageBusBuilder : IHasPostConfigurationActions, ISerializationBui
 
     public IList<Action<IServiceCollection>> PostConfigurationActions { get; } = [];
 
+    protected IList<IHasPostConfigurationActions> ConsumerPostConfigurationActions { get; } = [];
```
zarusz, line 29: I wonder if we could consider two other approaches:

1) On the MessageBusBuilder collect all the ConsumerBuilder that are added and later iterate through them in the `GetPostConfigurationActions()`

2) Let the AbstractConsumerBuilder.PostConfigurationActions act as the proxy to MessageBusBuilder.PostConfigurationActions - so whenever a post action is added to the consumer builder it results in the bus builder actions collection. Unless order of the execution of actions matters (e.g first consumer actions, then bus actions).

src/SlimMessageBus.Host.CircuitBreaker.HealthCheck/HealthCheckBackgroundService.cs
```diff
@@ -0,0 +1,125 @@
...
+        _tagStatus = new Dictionary<string, HealthStatus>();
+    }
+
+    public IReadOnlyDictionary<string, HealthStatus> TagStatus => _tagStatus;
+
+    public async Task PublishAsync(HealthReport report, CancellationToken cancellationToken)
+    {
+        var linkedToken = CancellationTokenSource.CreateLinkedTokenSource(_cancellationTokenSource.Token, cancellationToken).Token;
+
+        await _semaphore.WaitAsync(linkedToken);
```
zarusz, line 26: pass CT

src/SlimMessageBus.Host/Consumer/AbstractConsumer.cs
```diff
@@ -2,66 +2,122 @@
 
 public abstract class AbstractConsumer : IAsyncDisposable, IConsumerControl
```
zarusz, line 3: Could we invert the resposibilities here?

As it is the SlimMessageBus.Host knows about circuit breakers while its an opt-in plugin. I sense a leakage of concerns here (leakage of concerns).

What if the circuit breaker plugin (and its background task) it could traverse the IMasterBus and its Consumers to find the ones with matching tags and either start or stop them depending on configuration?

That way we contain the circuit breaker logic in the plugin assembly, and the .Host layer is free of that logic.

docs/intro.t.md
EtherZa: Updated and pushed.

src/SlimMessageBus.Host.Configuration/Builders/MessageBusBuilder.cs
EtherZa: 1. Can be done fairly simply but will result in the same touch points as `ConsumerPostConfigurationActions.Add(b);`. Each consumer will need to be added to a collection on the `MessageBusBuilder`.
2. This is a better approach but requires changing the `AbstractConsumerBuilder's` constructor to receive the `MessageBusBuilder` instance instead of `MessageBusSettings`. The constructor could then extract the `Settings` and continue processing as before while referencing the `MessageBusBuilder.PostConfigurationActions` property. The change spirals pretty quickly (impacting consumers, handlers and tests) but with small changes for the most part. More effort will be required to change `MessageBusMock` but is doable.

src/SlimMessageBus.Host/Consumer/AbstractConsumer.cs
EtherZa: I originally had a look at this approach, but if a circuit breaker traverses `IMasterBus` as you suggest:

1.  it will be ignorant to any other status changes made by other workflows and then may unwittingly reverse an intended configuration change
2. be side stepped by toggling the `IConsumerControl` methods directly via `MessageBusBase.Consumers` (unable to start with circuit breakers already in the loop)
3. require a centralised co-ordinator to aggregate state for all circuit breakers (shifting complexity but still resulting in an inferior experience)

The alternate approach that I investigated was to create a shim/proxy that intercepts the `Consumers` property in `MessageBusBase`. It could then trip/open the circuits as required, but would need to change the collection to be that of `IConsumerControl` so that a proxy could intercept the methods. This approach then cuts off all access to the underlying consumers.

ie.
```cs
protected void AddConsumer(AbstractConsumer consumer) => _consumers.Add(new CircuitBreakerProxy(consumer));
```

In the end the supplied implementation rendered the best results at the expense of being more invasive than ideal, but please let me know if you have another approach.

zarusz: I feel pretty strongly we need to go for externalization of this logic outside of the `.Host` layer and into the plugin layer which needs to coordinate consumers and maintain the Started/Stopped state based on the respective circuit state (I see the plugin acting as an adapter).

If anything we could add a factory method to be able to wrap these consumers into some proxies (from the circuit plugin) - or another smart design.

I am less concerned that someone will use `IConsumerControl` explicitly when using a circuit breaker. And even if so, then if the consumers were never started (or are stopped imperatively) the breaker logic could detect that and could withhold from starting such consumers if the breaker is on again.

The one thing that might get tricky with externalized logic is to maintain identity (or a pointer) to the consumer instance, but if anything we could (assign some identity to these consumers to find them easily by ID). However recalling vaguely the AbstractConsumer instances should remain (not being disposed) between Start/Stop cycles.

I am open to prototyping this logic on the side together if that helps express my intent.

EtherZa: Okay, I'll need to dig into it a little bit, but I am not sure when I will be able to do so. I have some pretty tight deadlines over the next 3 months which are going to leave little time unfortunately.

I think that I have a fair grasp of what you are after, but there are some immediate issues that will need to be overcome as you have described:
1. `MessageBusBase` does create/dispose the consumers between Start/Stop cycles.
2. There is no reference between configuration and consumer. I added this in the current PR but making it externally visible is not ideal as the objects are not immutable.
3. `MessageBusHostedService` can be replaced to manage the running state, but due to 1, there is no interception point to stop a consumer from starting if the circuit has already been tripped.

zarusz:
@EtherZa  would it be possible to either:
1. Rebase on release/v3, resolve few conflicts and change PR target to release/v3? I could merge as is then and refactor based on my earlier ask (shift some responsibilities).
2. Incorporate my earlier feedback. I understand you might have other priorities, hence option 1.

EtherZa:
@zarusz I have rebased the change on `release/v3`. If you have the appetite to do so, you are more than welcome to jump in and make the changes. Work committments are still a little on the heavy side at the moment.

PR #283, opened 2024-07-17
Sql outbox - suppress transactions for all OutboxSendingTask operation

EtherZa:
Suppress transactions for all queries related to the `OutboxSendingTask` and `OutboxLockRenewalTimer` classes. 

Promotion to distributed transactions is being experienced with the #282 build and should not occur as locks are managed outside of transactions.

```
[xUnit.net 00:01:07.39] [08:10:42 DBG] SlimMessageBus.Host.Outbox.OutboxForwardingPublishInterceptor Forwarding published message of type CustomerCreatedEvent to the outbox
[xUnit.net 00:01:07.39] [08:10:42 ERR] Microsoft.EntityFrameworkCore.Update An exception occurred in the database while saving changes for context type 'SlimMessageBus.Host.Outbox.DbContext.Test.DataAccess.CustomerContext'.
[xUnit.net 00:01:07.39] System.PlatformNotSupportedException: This platform does not support distributed transactions.
[xUnit.net 00:01:07.39] at System.Transactions.Oletx.OletxTransactionManager.CreateTransaction(TransactionOptions options)
[xUnit.net 00:01:07.39] at System.Transactions.TransactionStatePromoted.EnterState(InternalTransaction tx)
[xUnit.net 00:01:07.39] at System.Transactions.EnlistableStates.Promote(InternalTransaction tx)
[xUnit.net 00:01:07.39] at System.Transactions.Transaction.Promote()
```
---
I have also increased the timeout of the `KafkaMessageBusIt.BasicReqResp` test from 10 seconds from 5 seconds. Topology provision on local containers is failing to complete within the original timeout.

### DIFF
--- a/src/SlimMessageBus.Host.Outbox.Sql/SqlOutboxRepository.cs
+++ b/src/SlimMessageBus.Host.Outbox.Sql/SqlOutboxRepository.cs
@@ -4,12 +4,14 @@ public class SqlOutboxRepository : CommonSqlRepository, ISqlOutboxRepository
 {
     private readonly SqlOutboxTemplate _sqlTemplate;
     private readonly JsonSerializerOptions _jsonOptions;
+    private readonly string _connectionString;
 
     protected SqlOutboxSettings Settings { get; }
 
     public SqlOutboxRepository(ILogger<SqlOutboxRepository> logger, SqlOutboxSettings settings, SqlOutboxTemplate sqlOutboxTemplate, SqlConnection connection, ISqlTransactionService transactionService)
         : base(logger, settings.SqlSettings, connection, transactionService)
     {
+        _connectionString = connection.ConnectionString;
         _sqlTemplate = sqlOutboxTemplate;
         _jsonOptions = new();
         _jsonOptions.Converters.Add(new ObjectToInferredTypesConverter());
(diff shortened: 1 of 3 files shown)


src/SlimMessageBus.Host.Outbox.Sql/SqlOutboxRepository.cs
```diff
@@ -41,17 +42,20 @@ public async virtual Task Save(OutboxMessage message, CancellationToken token)
...
     }
 
-    public async Task<IReadOnlyCollection<OutboxMessage>> LockAndSelect(string instanceId, int batchSize, bool tableLock, TimeSpan lockDuration, CancellationToken token)
+    public async Task<IReadOnlyCollection<OutboxMessage>> LockAndSelect(string instanceId, int batchSize, bool tableLock, TimeSpan lockDuration, CancellationToken cancellationToken)
     {
-        await EnsureConnection();
+        // use connection outside of any transaction
 
-        using var cmd = CreateCommand();
+        using var scope = new TransactionScope(TransactionScopeOption.Suppress, TransactionScopeAsyncFlowOption.Enabled);
```
zarusz, line 49: Help me understand here please. We specify the option Suppress to let the commands run without the transaction scope. However why does it matter if the outbox loop runs (to mu understanding) without a spanning transaction scope or transaction?

Sorry just trying to understand the context here and it this is truly needed.

EtherZa: Quite simply, it is because I am a moron that misread the exception message. It's the `OutboxForwardingPublishInterceptor` that was escalating to a DT and not the `OutboxSendTask`. Oh boy...

The outbox loop must definitely run outside of any shared transaction due to how it handles concurrency, but that is moot due to it running as a parallel task.

-Sorry for wasting your time :(

zarusz:
All good @EtherZa. Thanks for looking into this tough.

PR #298, opened 2024-08-31
[Host.Outbox] Execute permission is required for Outbox #297

### LINKED ISSUE #297: [Host.Outbox] Execute permission is required for Outbox
Hey,

When trying to use Outbox feature, we receive the following error: `The EXECUTE permission was denied on the object 'OutboxId', database 'SQL-DB-SAMPLE', schema 'dbo'.` 

My understanding that EXEC permission is required for the custom `OutboxId` type, and that error could be fixed with something like `GRANT EXEC ON TYPE::[dbo].[OutboxId] TO [User]` but it's an issue to get such permission in secured environments and I wonder if it's possible to use Outbox without the custom type.

Thanks!

zarusz:
Fixes #297.

### DIFF
--- a/src/Host.Plugin.Properties.xml
+++ b/src/Host.Plugin.Properties.xml
@@ -4,7 +4,7 @@
   <Import Project="Common.NuGet.Properties.xml" />
 
   <PropertyGroup>
-    <Version>2.5.1</Version>
+    <Version>2.5.2-rc1</Version>
   </PropertyGroup>
 
 </Project>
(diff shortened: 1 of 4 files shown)


zarusz:
@EtherZa do you have a moment to review this change?
It removed the need for the custom SQL type.

EtherZa:
Hi @zarusz. That should work but will come with a couple caveats
1. There is obviously more overhead with this approach and performance will be impacted but it will probably be negligable due to the numher of GUID's sent in a single update (max 50 if memory serves me).
2. It increases the SQL compat level from 100 (SQL 2008) to 130 (SQL 2016). SQL Server 2016 is currently the earliest version with official support; so it is probably not a concern (SQL 2014 was EOL last month).

zarusz:
Thanks for your input. The DataTable object also had some allocation footprint. The batch size is going to have some practical limitations  but can't see why we'd need a huge number. 
Compatibility due to the used built in function is a fair point but as you said folks likely migrated.

PR #315, opened 2024-10-03
[Host.Outbox] Support for UUIDv7, different message keys and refactor

### LINKED ISSUE #311: [Host.Outbox]  Allow implementation OutboxRepository for a specific type PK
I want to create outbox repository for Postgres, and I need the following features to create tables as PARTITION:
 
  - generate monotonically increasing PK
  - choose type for PK ( int64,  [ULID](https://github.com/Cysharp/Ulid),  UUID v7 )

for example:

``` csharp
public interface IOutboxRepository<TOutboxKey>  // <==  specific type
{
    Task<TOutboxKey> GenerateId(CancellationToken cancellationToken);  // <= added method
    Task Save(OutboxMessage<TOutboxKey> message, CancellationToken token);
``` 

The current implementation in the base class does not allow to do this.

zarusz:
Several changes to `SlimMessageBus.Host.Outbox.*` in preparation for onboarding other storage providers:
- `IGuidGenerator` in `SlimMessageBus.Host`
- Default `GuidGenerator` registered in DI
- Ability to register custom UUIDv7 implementation [based on net9.0](https://steven-giesel.com/blogPost/ea42a518-4d8b-4e08-8f73-e542bdd3b983) or 3rd party library before net9.0
- In the `.Outbox.Sql` layer the ability in the `SqlOutboxSettings` to
  - specify client ID generation or database ID generation (GUIDs),
  - specify `GuidGeneratorType` in the MSDI or `GuidGenerator` instance for the client ID generation.
- Renamed `SlimMessageBus.Host.Outbox.DbContext` to `SlimMessageBus.Host.Outbox.Sql.DbContext` - to express its bound to MSSQL
- I've moved the "system fields" from `OutboxMessage` down to the .Outbox.Sql layer into a specialized type `SqlOutboxMessage`.
- The PK for an outbox message is now a generic, and so we should allow for other types than Guid as the PK. `.Outbox.Sql` will still use the `Guid` for backward compatibility

Closes #311


src/SlimMessageBus.Host.Outbox/Interceptors/OutboxForwardingPublishInterceptor.cs
```diff
@@ -70,6 +75,8 @@ public async Task OnHandle(T message, Func<Task> next, IProducerContext context)
         // Add message to the database, do not call next()
         var outboxMessage = new OutboxMessage
         {
+            Id = guidGenerator.NewGuid(),
```
dundich, line 78: I am against generating the Id (+timestamp) here, it should be in the repository, since it is closely related to the DB and the selection strategy should be in one place.

src/SlimMessageBus.Host.Outbox/Repositories/OutboxMessage.cs
dundich, line 1: I think you should delete all system fields that are used only in sql and tests.


```csharp
public class OutboxMessage
{
    public string BusName { get; set; }
    public string MessageType { get; set; }
    public byte[] MessagePayload { get; set; }
    public string Path { get; set; }
    public IDictionary<string, object> Headers { get; set; }
    public string InstanceId { get; set; }
    
   //*** system fields ***
    
    // public Guid Id { get; set; }
    // public DateTime Timestamp { get; set; }
    //  public string LockInstanceId { get; set; } = string.Empty;
    //  public DateTime? LockExpiresOn { get; set; } = null;
    //  public int DeliveryAttempt { get; set; } = 0;
    //  public bool DeliveryComplete { get; set; } = false;
    //  public bool DeliveryAborted { get; set; } = false;
}
```

// in repository

    public async virtual Task Save(OutboxMessage message, CancellationToken token)
    {
            // direct in sql param
                Id = reader.GetGuid(idOrdinal),
                Timestamp = reader.GetDateTime(timestampOrdinal),

src/SlimMessageBus.Host.Outbox/Interceptors/OutboxForwardingPublishInterceptor.cs
zarusz: Oh yes, that would be a good idea. The repo implementation (and chosen strategy via setting) could know how to generate the IDs. Just notice this is a draft PR to get something quick together for now :)

src/SlimMessageBus.Host.Outbox/Repositories/OutboxMessage.cs
zarusz: Okay I see, yeah we could hide some properties from the public contract/interface. Yet again properties like DeliveryComplete  are relevant for SQL and non SQL storage...

zarusz: So rereading this again and making sure we are on the same page - the `OutboxMessage` is the actual entity (regardless of the SQL or not) its the actual entity that is passed to the IOutboxRepository. The ID/timestamp could be owned by the implementation (SqlOutboxRepository) but I don't see a need to create another layer/wrapper type on top of OutboxMessage nor to remove the "system" properties - they are part of the stored record.

dundich: You are right "system" properties - they are part of the stored record!

I mean that a clean input object (without sys fields) should be passed to the Save method

let's call it more correctly

```csharp
public class OutboxMessageInput
{
    public string BusName { get; set; }
    public string MessageType { get; set; }
    public byte[] MessagePayload { get; set; }
    public string Path { get; set; }
    public IDictionary<string, object> Headers { get; set; }
    public string InstanceId { get; set; }
}


  ~ Task<Guid | OutboxMessage> Save(OutboxMessageInput message)
```

A contract requires unambiguity, so it requires a different type of input class that does not allow for different interpretations.

I don't think there will be any dispute here?

dundich: > the OutboxMessage is the actual entity (regardless of the SQL or not)

Let's say yes, but the code doesn't show how they are used outside of sql, which begs the question, why are we reading this sys context and passing it outside?

Maybe later this will come in handy to change behavior, but for now why race them around the network (sys fields).

zarusz:
@dundich  here is another iteration:

- `IOutboxMessageFactory` to be able to create messages (and store), Id/Timestamp assignment happens in the implementation. It  takes only the minimal params in and returns the outbox message Id.
- `IOutboxMessageRepository` does not have the `Save` method anymore - in favor of the above
- `SqlOutboxMessageRepository` is the implementation of the above two interfaces, currently tied to MSSQL but could get a postgres dialect
- `SqlDialect` - see this, it will allow to specify the dialect, just added Postgres

There is some more work needed to drive by the OutboxSettings configuration to inform what a concrete implementation of `IGuidGenerator` to resolve from MSDI - this will be the main bit that will allow to generate the uuidv7 for you (either net9 or plugging a 3rd party lib like [NewId](https://github.com/phatboyg/NewId) for < net9.

dundich:
I agree it is worth defining the API for connecting Outbox strategies. 

But before Api I think it is necessary to define possible interception points for working with OutboxFlow.  So far I see them in three places:

## interception points

1)  current (classic) mode :  first outbox -> transport    (https://microservices.io/patterns/data/transactional-outbox.html)
2)  or optimistic mode     : first transport -> if error -> outbox (https://github.com/zarusz/SlimMessageBus/issues/316)
3)  CircuitBreaker -  if no check -> outbox   https://github.com/zarusz/SlimMessageBus/pull/282

They should probably be set up through configuration.

## hybrid outbox ?

like as (https://github.com/zarusz/SlimMessageBus/blob/master/docs/provider_hybrid.md) for (mssql, pg, redis)

but imho overhead - maybe it's not worth it yet ?

## Describe the prototype connection API

like as https://github.com/zarusz/SlimMessageBus/pull/282#issue-2412572427


ps

Perhaps then it will become clearer whether we need additional abstractions.

zarusz:
@dundich all great ideas, but that needs to be done in stages. 
I don't want this PR to snow ball into many new features.

The intent of this PR is to introduce a way to inject the Guid generator and connect the existing time provider interface into outbox. This is to enable your need for UUIDv7.

Any further enhancements need to come after that.

Next stages could include:
- enabling NET9 rc targeting (started #318)
- merging v3 of SMB (with #223) to master and releasing stable v3 (want to do that soon)
- enabling Postgress as dialect (feel free to pick this up)
- the hybrid outbox and circuit breakers (feel free to pick this up)

Let me know if you have other priorities in mind.

zarusz:
As for the hybrid outbox my thinking is this:
- Any other SQL database (e.g. Pg, MsSql, MySql) could simply be a dialect setting (on the SlimMessageBus.Host.Outbox.Sql layer with necessary adjustments).
- When we need to go to leveraging let say Redis for storing outbox message that it simply be another layer I am thinking (`SlimMessageBus.Host.Outbox.Redis`). 

Also, the idea with `SlimMessageBus.Host.Outbox` layer is that it should contain enough generic logic for any type of outbox storage used.

Not sure that you'd need a hybrid outbox with both some messages routed to SQL outbox, others to Redis Outbox.
Unless this is what you have good use case for and need? If so outbox-es could be named (yes like in the hybrid) and then routed from the publish outbox interceptor to the right outbox.

Again, we need to do this in stages :)

dundich:
I agree with all of this - the elephant needs to be eaten in pieces.
As for the hybrid outbox - unnecessary engineering, I don't think it will be useful.

As for the dialect, I didn't quite catch the point. I thought it was enough to register my IPgSqlRepository implementation via config... and why I need a dialect, I don't understand yet.....

 it seems to me that you want to put everything in one SlimMessageBus.Host.Outbox.Sql  or am I wrong?

zarusz:
> As for the dialect, I didn't quite catch the point. I thought it was enough to register my IPgSqlRepository implementation via config... and why I need a dialect, I don't understand yet.....

> it seems to me that you want to put everything in one SlimMessageBus.Host.Outbox.Sql or am I wrong?

Oh yes I see we would need a new project `SlimMessageBus.Host.Outbox.PqSql` that references [NpgSql](https://www.nuget.org/packages/Npgsql/), that injects a new `ISqlOutboxMessageRepository` (or `IOutboxMessageRepository`) implementation. I thought we could use the SqlClient already present and just via a SqlDialect property control which SQL scripts are used for MSSQL or Postgres - but that is not the case.

zarusz:
@dundich I am ready with all my changes. Hopefully, this will give a baseline for your Postgress Outbox work (although some more refactoring might be needed). Certainly, it will allow for Guid generation strategies e.g. UUIDv7.

dundich:
Refactoring - need to understand how best to reorganize two projects with mssql:

1. Outbox.Sql
2. Sql.Common

https://github.com/zarusz/SlimMessageBus/pull/315#issuecomment-2395335763
ps. I think a separate ticket is required for this ...

dundich:
> @dundich I am ready with all my changes. Hopefully, this will give a baseline for your Postgress Outbox work (although some more refactoring might be needed). Certainly, it will allow for Guid generation strategies e.g. UUIDv7.

There is one point - since the contract has been rewritten, now id generation may come with a response from Pg or Mssql, the generator may be useless here  ([uid_generate_v1mc()](https://www.postgresql.org/docs/9.4/uuid-ossp.html)).

zarusz:
@dundich  I am debating whether not to refactor this `Outbox.Sql` and `Outbox.DbContext` even further
The `EntityFrameworkCore` could act as the database abstraction to any type of SQL database (MS SQL, and Postgres) by bringing in 
- [Npgsql.EntityFrameworkCore.PostgreSQL](https://www.nuget.org/packages/npgsql.entityframeworkcore.postgresql/) and https://www.npgsql.org/efcore/?tabs=onconfiguring for Postgres
- [Microsoft.EntityFrameworkCore.SqlServer](https://www.nuget.org/packages/Microsoft.EntityFrameworkCore.SqlServer) for MSSQL

There are still a few considerations around managing transactions
- currently, the SqlTransaction from SqlClient is used in one of the interceptors as a way to wrap unit of work in such a transaction ([.UseSqlTransaction()](https://github.com/zarusz/SlimMessageBus/blob/master/docs/plugin_outbox.md#introduction))
- the TransactionScope approach should still work for the above approach ([.UseTransactionScope()](https://github.com/zarusz/SlimMessageBus/blob/master/docs/plugin_outbox.md#introduction))
- we could leverage the [_dbContext.Database.BeginTransactionAsync()](https://learn.microsoft.com/en-us/dotnet/api/system.data.common.dbconnection.begintransactionasync?view=net-8.0) to be a transaction abstraction too (replacing the first approach)

The only caveat is if we let EF Core to manage the outbox messages it might end up being slower than the current T-SQL implementation.

cc: @EtherZa (in case wanted to weigh in)

EtherZa:
@zarusz I would be apprehensive to create a single plugin that supports multiple database providers. 

There are slight variations in the dialects that would result in a number of conditionals (non-clustered indexes in sql but not in postgres or mysql, select top vs limit, etc). Further to this, there are features that can be leveraged in some providers that are not available in others (postgres [notficiations](https://www.postgresql.org/docs/current/sql-notify.html)). As such, you will get a far better implementation with a tailored solution.

As for using EF; you run the risk of pinning the EF version which would be a deal breaker for me personally. It would also reduce the feature set to the lowest common denominator.

I agree with @dundich that the implementation/DTO could be cleaned up. The provider specific implementation can be hidden from `SlimMessageBus.Host.Outbox` - this could even include the type of id being used (performance would be improved in the MS SQL implementation if a sequence were to be used instead of a GUID).

----
As a side note (and definitely scope creep for this discussion); the current implementation could be vastly improved if the message processing were to be partitioned by path (and subscription for ASB). Processing by partition (round-robin, instead of purely by sequence) would ensure that multiple queues are populated instead of just one working through a backlog.

ie. In the workflow A => B, where A enqueues a message for B; if 100k messages are outboxed for the A queue, the current implementation will sequentially run through the messages adding all 100k to A before starting to add anything to the B queue. The B consumers then sit idle while A works its way through the backlog.

I only bring this up as it could be benefitial to keep in mind if refactoring is planned.

dundich:
I would like to join right now, but I am busy with other things right now and so I will keep it short

I am also against linking to EF. I want to use PG features such as: partitions,  [skip-locked](https://www.2ndquadrant.com/en/blog/what-is-select-skip-locked-for-in-postgresql-9-5/)

I would focus on contract to connect different implementations and customize them.

1) Definition of DTOs (sys fields)
2) Ability to save a packet (e.g. for buffered flush)
3) Configuration (base, specific for implementation)  
4) etc.

dundich:
Again, I think the problem lies in the name - repository.

It is clear that a repository should be as simple as possible, hence the desire to have one BLL with single settings. I think this is a mistake. 

We should allow each vendor to have their own properties and define Outbox logic independently, just like Buses have their own specifics..

zarusz:
@dundich  and @EtherZa  thanks for your perspectives and ideas.

The idea with EF Core serving as an abstraction would be an addition to ensure broader databases coverage. So let say we would still have a MSSQL (like today) and PG specific plugin (hopefully built by @dundich), but any other db would fallback to a slower EF implementation. I share the same over abstraction and lower performance concern in contrast when comparing EF approach vs optimized Outbox.Sql or PQ.

Existing ones that we'd preserve in v3.0.0
```
SlimMessageBus.Host.Outbox (shared abstracted outbox logic with IOutboxMessageRepository)
SlimMessageBus.Host.Outbox.Sql (specific to MSSQL)
SlimMessageBus.Host.Outbox.Sql.DbContext (depends on SlimMessageBus.Host.Outbox.Sql, renamed from SlimMessageBus.Host.Outbox.DbContext
```

The new ones I can foresee:
```
SlimMessageBus.Host.Outbox.Postgres (specific to Postgres, ideally dependant on SlimMessageBus.Host.Outbox to drive re-use)
SlimMessageBus.Host.Outbox.DbContext (a new one that uses EntityFrameworkCore as the abstraction, slower)
SlimMessageBus.Host.Outbox.Memory (an in memory buffer to postpone the publish of messages until unit of work finish / tx.Commit)
```

@EtherZa  the partitioning feature on MSSQL is a good one, I will capture as another feature request.

@dundich again, the repository is a way to delegate the db specific SQL dialect (e.g. Outbox.Sql / Outbox.Pg) and hopefully for the db specialized plugins to still leverage the common logic within Host.Outbox. We could rename to OutboxService or something if that sits better. Now, we tried to move the PK out of the entity to allow for provider specific Key types, I could still try to revive that. My stance on the abstraction in `Host.Outbox` - if its is not good enough for Pg implementation this is where I think (once you get there) make a decision to refactor the `Host.Outbox` layer some more or part ways from it and make the Pg completely custom (I don't think there would be 0 reuse between MSQL and Pg).

zarusz:
@dundich, I think I am now done with my changes. 

* I've moved the "system fields" down to the `.Outbox.Sql` layer into a specialized type `SqlOutboxMessage`.
* The PK is now a generic, and so we should allow for other types than Guid as the PK. `.Outbox.Sql` will still use the Guid
* But `.Outbox.Sql` will be able to choose the Guid generation strategy - I've moved the strategy settings over to that layer too. If you'd need something similar in Postgress, bring it down to `.Outbox`, but I sense you will go down a different route more optimal towards PG.
* I will have my other exploratory feature branch open (in memory outbox + abstraction using EntityFrameworkCore) and perhaps evolve it depending on time and priority.

Happy to merge, unless you see anything out of line.

dundich:
pls look at small fix for uuidv7 - https://github.com/zarusz/SlimMessageBus/pull/325

zarusz:
Yes, good idea! Merged.

zarusz:
@dundich I managed to stabilize the Outbox tests. There is more work to be done in the near future but will park this for now under #329 to allow you to go forward with the Postgress work. I have few other priorities too emerging now.

Issue #347, opened 2024-11-28
[Host] IConsumerErrorHandler usage limited when using per message scopes


EtherZa:
When per message scoping is enabled and an exception is thrown in the messaging pipeline; if a custom imeplementation of `IConsumerErrorHandler` is available, both the instnace of `IConsumerErrorHandler` and the `retry` function supplied to `IConsumerErrorHandler.OnHandleErorr()` are created under the original scope. This limits the functionality of the retry mechanism  when transactions/scoped dependencies are in use.

An use case is when using a scope registered `DbContext` that experiences a transient error while in a transaction. Any subsequent retry will reuse the existing `DbContent` connection which is now in a failed state.

It would be ideal to only create the message scope when `retry()`/`DoHandleInternal()` is called, but the `consumerInvoker` instance that is made available to both `IConsumerErrorHandler` and `IConsumerContext`  is dependent on the scope.

EtherZa:
`IConsumerContext` is also dependent on a scoped instance of `IMessasgeBus` (`MessageBusProxy`).

zarusz:
Are you planning a PR improve this? 
If this something that is a priority for your project?

EtherZa:
To be honest, there's not a lot that I can do to improve it without introducing breaking changes.

The basic flow at the moment is:
1. create a `IMessageScope`
2. create `MessageBusProxy` using scope
3. create `consumerInstance` using scope
4. create `IConsumerContext` using scope with `MesageBusProxy` + `consumerInstance`
5. call `DoHandleInternal()`, passing the scoped `IConsumerContext` and `MessageBusProxy`
6. on exception: call `DoHandleError()` passing int the same `MessageBusProxy`, `IConsumerContext` and `IMessageScope`
7. Check for an isntance of `IConsumerErrorHandler` and recreate the steps in `DoHandleInternal()` to be co-ordinated by the `IConsumerErrorHandler` instance.

Due to this implementation being wrapped by a single `IMessageScope`, the `IConsumerErrorHandler` and `retry()` instances executes under the same scope as the original execution.

**Possible solutions:**
1. Run through steps 1 to 5 as above and on an error check if there is an instance of `IConsumerErrorHandler` available. If so, create a new scope and execute `OnHandleError()` on the instance. `IConsumerErrorHandler` allows for multiple retry executions but due to the wrapped scope, the exception will just be deferred to the second exexcution.
[...]

PR #354, opened 2024-12-29
[Host] Refactor 'Retry' delegate in IConsumerErrorHandler as response type

EtherZa:
While the discussed 'pipeline reimagination' was a little too ambitious for the time that I have available; I have refactored the implementation to extract the `retry` delegate from `IConsumerErrorHandler.OnHandleError` and to rather include it as a response value. This does introduce a breaking change, but allows the handler to perform all of the previous logic albeit with a small change for retry.

The `OnHandlerError` method signature therefore changes from:
```cs
Task<ConsumerErrorHandlerResult> OnHandleError(T message, Func<Task<object>> retry, IConsumerContext consumerContext, Exception exception);
```
to:
```cs
// where `attempts` is the execution count
Task<ConsumerErrorHandlerResult> OnHandleError(T message, IConsumerContext consumerContext, Exception exception, int attempts);
```
In order to retry the execution (with a new scope as requried), the implemented method must return `ConsumerErrorHandlerResult.Retry`. Further to this, abstract implmentations of `IConsumerErrorHandler` have been added with methods to support `Failure()`, `Success(object? response = null)` and `Retry()` for a cleaner experience (similar to `return Ok();` with asp.net controllers).

Example:
```cs
public class RetryHandler<T> : ConsumerErrorHandler<T>
{
    private static readonly Random _random = new();

    public override async Task<ConsumerErrorHandlerResult> OnHandleError(T message, IConsumerContext consumerContext, Exception exception, int attempts)
    {
        if (attempts <= 3)
        {
            var delay = (attempts * 1000) + (_random.Next(1000) - 500);

            await Task.Delay(delay, consumerContext.CancellationToken);
            return Retry();
        }

        return Failure();
    }
}
```


docs/intro.t.md
```diff
@@ -1059,14 +1059,17 @@ public class LoggingConsumerInterceptor<TMessage> : IConsumerInterceptor<TMessag
...
-> The `retry()` parameter allows the message processing pipeline, including consumer interceptors, to retry processing when transient errors occur and retries are desired.
+The returned `ConsumerErrorHandlerResult` object is used to override the execution for the remainder of the execution pipeline.
+| Result  | Description |
+|---------|-------------|
+| Failure | The message failed to be processed and should be returned to the queue |
+| Success | The pipeline must treat the message as having been processed successfully |
+| SuccessWithResponse | The pipeline to treat the messagage as having been processed successfully, returning the response to the request/response invocation ([IRequestResponseBus<T>](../src/SlimMessageBus/RequestResponse/IRequestResponseBus.cs)) |
+| Retry   | Execute the pipeline again (any delay/jitter should be applied before returning from method) |
 
 To enable SMB to recognize the error handler, it must be registered within the Microsoft Dependency Injection (MSDI) framework:
```
zarusz, line 1075: @EtherZa  could you also add the sample error handler from the PR into the into.t.md?
I think people will find it handy to have a starting sample.

zarusz: Also please make a note about the fact the message scope will be re-created with each retry. Its not apparent when reading this description.

zarusz:
This is an excellent and pragmatic approach to improving the error-handling pipeline while achieving the message scope recreation as discussed in issue #347. The implementation is mostly backward-compatible and could potentially outperform the `Func<Task> next` delegate approach.

I've suggested a few minor adjustments to the documentation for clarity and completeness.

Once those are addressed, I’ll be happy to approve the PR. Great work!

docs/intro.t.md
EtherZa: Sample and note added.

zarusz:
It's been released under v3 rc:
https://www.nuget.org/packages/SlimMessageBus.Host.AzureServiceBus/3.0.0-rc901

Issue #356, opened 2024-12-31
[Host] Abandon/dead letter message


EtherZa:
SMB does not provide an out of the box way to dead letter messages for the transports that support it. It is currently possible for RabbitMQ through a nack in a `IConsumerErrorHandler`, but not for Azure Service Bus.

ASB messages that throw an exception will be retried n times before finally being sent to the DLQ. 

If an exception can be determined to be non-transient (and therefore will not be resolved in a subsequent execution), it would be beneficial to send the message to the DLQ immediately.

zarusz:
Yes, we could add this as an asb transport specific feature. Perhaps even unify across all transports (but not sure about if it's worth unifying).

zarusz:
What would be good is also gave the ability to set application properties (headers) from the error handler or consumer when message is abandoned on the asb transports. This helps to convey partially completed work and have it available for the next message retry.

zarusz:
Closed via #357.

Also see another feature request: #358

PR #357, opened 2025-01-01
[Host] Abandon/dead letter (#356)

EtherZa:
Added `Abandon` response to `ConsumerErrorHandlerResult` to be used by supported transports to dead letter the message. Most transports don't support the feature without some form of customised flow, but ASB and RabbitMQ both do.

***Azure Service Bus***
Implemented without restriction. Abandoned messages will be sent directly to the DLQ, failed messages will be retried until their maximum delivery count has been exceeded.

***RabbitMQ***
I don't have any experience with RabbitMQ, so I may have missed something; but the bus does not appear to limit how many times a message will be delivered if it is Nacked with requeue = true. As such, changing the code to fail with a requeue would be a dangerous breaking change as it could lead to an inifinite recursion in message processing for exisiting SMB users (unless I have missed something). 

I have left the RabbitMq processing flow intact but with the expectation that a future change will provide tracking or some other safety net for the above scenario. Do you have a better option/insight into the RabbitMQ model?

***Other transports***
A `NotSupportedException` will be thrown if `IConsumerErrorHandler` returns with an `Abandon` state.

***Application headers***
With ASB currently being the major benefactor to this change, it becomes a bit of a moot point. ASB does not support header customisation when dead lettering a message (`DeadLetterReason` and `DeadLetterErrorDescription` are both set, but that is the extent of the customization permitted).

### DIFF
--- a/docs/intro.t.md
+++ b/docs/intro.t.md
@@ -1067,10 +1067,11 @@ The returned `ConsumerErrorHandlerResult` object is used to override the executi
 
 | Result              | Description                                                                                                                                                                                                                 |
 | ------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Abandon             | The message should be sent to the dead letter queue/exchange. **Not supported by all transports.**                                                                                                                          |
 | Failure             | The message failed to be processed and should be returned to the queue                                                                                                                                                      |
 | Success             | The pipeline must treat the message as having been processed successfully                                                                                                                                                   |
 | SuccessWithResponse | The pipeline to treat the message as having been processed successfully, returning the response to the request/response invocation ([IRequestResponseBus<T>](../src/SlimMessageBus/RequestResponse/IRequestResponseBus.cs)) |
-| Retry               | Re-create and execute the pipeline (including the message scope when using [per-message DI container scopes](#per-message-di-container-scope)) |
+| Retry               | Re-create and execute the pipeline (including the message scope when using [per-message DI container scopes](#per-message-di-container-scope))                                                                              |
 
 To enable SMB to recognize the error handler, it must be registered within the Microsoft Dependency Injection (MSDI) framework:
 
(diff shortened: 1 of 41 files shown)


EtherZa:
@zarusz the build action has not been triggered

src/SlimMessageBus.Host.AzureServiceBus/Consumer/AsbBaseConsumer.cs
```diff
@@ -160,29 +165,38 @@ protected async Task ProcessMessageAsyncInternal(ServiceBusReceivedMessage messa
...
+                await completeMessage(message, token).ConfigureAwait(false);
+                return;
+
+            case ProcessResult.Abandon:
+                Logger.LogError(r.Exception, "Dead letter message - Path: {Path}, SubscriptionName: {SubscriptionName}, SequenceNumber: {SequenceNumber}, DeliveryCount: {DeliveryCount}, MessageId: {MessageId}", TopicSubscription.Path, TopicSubscription.SubscriptionName, message.SequenceNumber, message.DeliveryCount, message.MessageId);
+                await deadLetterMessage(message, r.Exception?.GetType().Name ?? string.Empty, r.Exception?.Message ?? string.Empty, token).ConfigureAwait(false);
+                return;
+
+            case ProcessResult.Fail:
+                var messageProperties = new Dictionary<string, object>();
```
zarusz, line 187: So I suppose the passing additional message headers to the ASB will require to be designed and implemented in another feature?

docs/intro.t.md
```diff
@@ -1067,10 +1067,11 @@ The returned `ConsumerErrorHandlerResult` object is used to override the executi
...
 | Success             | The pipeline must treat the message as having been processed successfully                                                                                                                                                   |
 | SuccessWithResponse | The pipeline to treat the message as having been processed successfully, returning the response to the request/response invocation ([IRequestResponseBus<T>](../src/SlimMessageBus/RequestResponse/IRequestResponseBus.cs)) |
-| Retry               | Re-create and execute the pipeline (including the message scope when using [per-message DI container scopes](#per-message-di-container-scope)) |
+| Retry               | Re-create and execute the pipeline (including the message scope when using [per-message DI container scopes](#per-message-di-container-scope))                                                                              |
 
 To enable SMB to recognize the error handler, it must be registered within the Microsoft Dependency Injection (MSDI) framework:
```
zarusz, line 1077: I think it would be beneficial to pass most of the PR description to here as its a good summary of overall error handling capabilities across transports in respect to the Abandon result.

zarusz:
Hi @EtherZa,

This implementation looks great! I’ve left a few minor comments for your review—one of them might be more appropriate for a separate PR to address as a standalone feature.

Regarding the GitHub build trigger issue, I’m not entirely sure what’s happening. Could you try force-pushing again to see if that resolves it?

Thanks for your contribution!

docs/intro.t.md
EtherZa: Added with details for `RequeueOnFailure` RabbitMQ configuration property (details below).

EtherZa:
> This implementation looks great! I’ve left a few minor comments for your review—one of them might be more appropriate for a separate PR to address as a standalone feature.

Docs have been updated. Message properties are not as straight forward as it appears. I've added details to the comment above. Prob best to have it as a separate change.

> Regarding the GitHub build trigger issue, I’m not entirely sure what’s happening. Could you try force-pushing again to see if that resolves it?

It's building again!

> Thanks for your contribution!

It's a pleasure :)

---
I have added a RabbitMQ configuration option `RequeueOnFailure()` that will issue a `Nack with requeue: true` on `Failure`. 

The default is still to `Nack` a message without a requeue but this can be changed by configuraing the switch per consumer. I have added a note to `intro.md` explicitly declaring that it is the developer's responsibility to `Abandon` all non-transient exceptions or face an inifinite loop with an example.

EtherZa:
Okay, I don't quite understand the build trigger. I force pushed; it built and failed. I resolved the issue and force pushed again; but there does not appear to be a subsequent build?!

zarusz:
@EtherZa I am happy to merge, however if possible perhaps you could increase coverage on some touched files to meet the coverage criteria?

EtherZa:
@zarusz I'll look at what I can add, but most of the infringements are due to reformatting (remove trailing space, etc) or gates on the transports that don't support `Abandon`.

ie.
```cs
if (r.Result == ProcessResult.Abandon)
{
    throw new NotSupportedException("Transport does not support abandoning messages");
}
```

Do you want a test for each of these?

zarusz:
Well, okay. Let's not bother. 

Please give me a thumbs up if you plan no more changes and then I will merge.

Issue #358, opened 2025-01-01
[Host.AzureServiceBus] Ability to set header when message has been is failed


zarusz:
Based on the discussion in #356 and #356, it would be beneficial to introduce the ability to set message headers (application properties) for Azure Service Bus when a message fails. This enhancement would allow users to:

- Preserve partial state information in the message headers during failure scenarios (either within the consumer or the consumer error handler).
- Ensure that this state is available for subsequent message retries, aiding in implementing consumer idempotency.

This feature would provide greater flexibility and reliability, especially in scenarios where preserving state across retries is critical.

PR #365, opened 2025-01-29
[Host.AzureServiceBus] Modify user properties on failure / Supply reason and description on dead-letter

EtherZa:
#358 Added support for suppliying modified user properties with using an ASB transport when using a custom `ServiceBusConsumerErrorHandler<T>` implemntation and `Failure()`.

```cs
public sealed class SampleConsumerErrorHandler<T> : ServiceBusConsumerErrorHandler<T>
{
    public override Task<ProcessResult> OnHandleError(T message, IConsumerContext consumerContext, Exception exception, int attempts)
    {
        var properties = new Dictionary<string, object>
        {
            { "Key", "value" },
            { "Attempts", attempts },
            { "SMB.Exception", exception.ToString().Substring(0, 1000) }
        };

        return Task.FromResult(Failure(properties));
    }
}
```
---
An option to supply a dead-letter reason/description when sending a message to the DLQ has also been included. If neither a reason nor description are supplied, it will fall back to the original implementation of exception type as the reason and exception message as the description.

```cs
public sealed class SampleConsumerErrorHandler<T> : ServiceBusConsumerErrorHandler<T>
{
    public override Task<ProcessResult> OnHandleError(T message, IConsumerContext consumerContext, Exception exception, int attempts)
    {
        return Task.FromResult(DeadLetter("reason", "description"));
    }
}
```
---
PS. It's great to see that v3 has escaped :)

### DIFF
--- a/docs/provider_azure_servicebus.md
+++ b/docs/provider_azure_servicebus.md
@@ -10,6 +10,8 @@ Please read the [Introduction](intro.md) before reading this provider documentat
   - [Default Subscription Name](#default-subscription-name)
   - [Consumer context](#consumer-context)
   - [Exception Handling for Consumers](#exception-handling-for-consumers)
+    - [DeadLetter: Application-Level Dead-Lettering](#deadletter-application-level-dead-lettering)
+    - [Failure: Modify Application Properties on Failure](#failure-modify-application-properties-on-failure)
   - [Transport Specific Settings](#transport-specific-settings)
 - [Request-Response Configuration](#request-response-configuration)
   - [Produce Request Messages](#produce-request-messages)
@@ -18,6 +20,7 @@ Please read the [Introduction](intro.md) before reading this provider documentat
 - [Topology Provisioning](#topology-provisioning)
   - [Validation of Topology](#validation-of-topology)
   - [Trigger Topology Provisioning](#trigger-topology-provisioning)
+
 
 ## Configuration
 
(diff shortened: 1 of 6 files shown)


docs/provider_azure_servicebus.md
```diff
@@ -193,12 +194,48 @@ This could be useful to extract the message's `CorrelationId` or `ApplicationPro
...
+{
+    public override Task<ProcessResult> OnHandleError(T message, IConsumerContext consumerContext, Exception exception, int attempts)
+    {
+        return Task.FromResult(DeadLetter("reason", "description"));
+    }
+}
+```
+
+#### Failure: Modify Application Properties on Failure
+An overload to `Failure(IReadOnlyDictionary<string, object)` is included to facilitate the modification of application properites on a failed message. This includes the `SMB.Exception` property should alternative detail be required.
```
zarusz, line 220: Try to add a new line after a heading. 

For markdown, I recommend visual studio code with https://marketplace.visualstudio.com/items?itemName=yzhang.markdown-all-in-one plugin. It reformats these for me and updates the table of content up top.

zarusz, line 220: Could you add another use case, somethign along the lines:

```
he properties may store partially processed data from a unit of work, which will be accessible to consumers through the `IConsumerContext` headers. This allows consumers to resume processing from specific points within the unit of work, ultimately supporting consumer idempotency - particularly in scenarios where distributed transactions are not feasible.
```

zarusz:
Thanks for another nice change that solves the #365.

Just a small suggestion on the docs, once addressed I am happy to merge.

Yes, v3 is out now :) If you need this faster I could release it under 3.0.1 or something.

docs/provider_azure_servicebus.md
EtherZa: I've updated the docs. Please let me know if you would prefer a different message/formatting/etc.

EtherZa:
Hi @zarusz. The feature would be certainly be useful, but it's no rush if you have other plans/etc for a delayed release.

EtherZa:
Hi @zarusz. Is it too late to change my mind on an early release?

I've been told that my team is producing around 100gb of logs a day and reckon that I can reduce it substantially by dynamically changing the logging behaviour on a failure.

zarusz:
Ok, let me push 3.0.1 today.

EtherZa:
Fantastic! Thank you very much.

zarusz:
It has been done: https://www.nuget.org/packages/SlimMessageBus.Host.AzureServiceBus/3.0.1

Issue #370, opened 2025-02-27
Pass topic name to message serializer


robertcoltheart:
I noticed that the Avro schema registry isn't supported, and I'd like to contribute to build in that support. 

The first part of this would be to pass the topic/path name to the message serializer, as this is required by the Avro serializer to send to the schema registry. 

I'll contribute a PR for this shortly.

zarusz:
hello @robertcoltheart, that sounds nice. 

Perhaps if you could approach the path/topic passing to the serializer by having an additional serialization interface to be implemented similar to the one I did for SQS
https://github.com/zarusz/SlimMessageBus/blob/9acab68c0e930d6c0aa48c25e319f88fb28f0d33/src/SlimMessageBus.Host.Serialization.SystemTextJson/JsonMessageSerializer.cs#L43

The SQS provider expected a string in the payload, so instead of serializing to byte[] the transport prefer the specialized string interface if one exists on the chosen serializer.

Of course, if we need to pass the path, and are forced to introduce a breaking change to IMessageSerializer, then it will be okay too.

robertcoltheart:
This is completed, will raise a new issue for adding in schema registry support.

PR #371, opened 2025-02-28
Pass path name to message serializers

robertcoltheart:
Fixes: #370 

This was a bigger change than I had anticipated, so apologies for that. I'm open to doing this a different way that is less intrusive. I considered using a `IMessageSerializer<string>` as per your suggestion, but this doesn't solve the problem of being able to pass the path name to the serializer during serialization. 

Another option would be to create a different serializer for each topic (for Avro), however it seems only singleton serializers are supported.

Happy to take your advice on the direction to go for this.


zarusz:
The number of changes looks reasonable. However, I do have some concerns about the additional allocation for `new MessageContext(path)`, which would occur with every message serialization. We might be able to optimize this by caching it outside of the lambdas (option 1a) or simply passing `path` as a parameter (option 1b).  

I'm also interested in your suggestion of having serializers per path (option 2a). This could allow us to keep the serialization interfaces backward-compatible, avoid additional allocations, and introduce an opt-in mechanism for different serialization strategies per path rather than per bus.  

The serializer instance is currently obtained here:  
[MessageBusBase.cs#L39](https://github.com/zarusz/SlimMessageBus/blob/9acab68c0e930d6c0aa48c25e319f88fb28f0d33/src/SlimMessageBus.Host/MessageBusBase.cs#L39).

Right now, it's pulled from MSDI and cached once per bus, effectively making it a singleton. That shouldn't be an issue, right? If we were to evolve `MessageBusBase` to look up `IMessageBusSerializer` per path, that could solve the problem while still keeping the serializer instances as singletons. So far, they’ve been singletons because there hasn’t been a strong need for anything different.  

I’d be happy to prototype option 2b to illustrate how a per-path serializer approach would look and provide a baseline for you to build upon in this PR.  

Let me know your thoughts on options 1a, 1b, and 2b. Do you have a strong preference?

robertcoltheart:
Yeah I like the idea of using a serializer per path, since the only serializer that will need more than one is the Avro serializer. This many changes for just one serializer felt off to me. Happy for you to prototype something - I'll have a look myself tonight as well.

zarusz:
@robertcoltheart, take a look at the draft PR #372. In this update, all serializer plugins now implement `IMessageSerializerProvider`, which enables retrieving `IMessageSerializer` instances per path. These instances are managed by the provider and reused accordingly. This change allows the Avro serializer to recognize the path during serialization.

While this introduced a significant number of changes, it was expected since serialization plugins now need to conform to the new interface, and the bus host runtime relies on it as the primary entry point for serialization.

I anticipate that we might need to provide the bus configuration to extract the paths for the serializer. This could enable it to eagerly load all per-path metadata, potentially avoiding the need for a `ConcurrentDictionary` when per-path serializer are requested at runtime. I may be thinking too far ahead, but I wanted to mention it in case you find it useful.

Let me know if you can iterate based on my feature branch. Happy to make further refinements.

robertcoltheart:
Thanks @zarusz this looks like it will do the trick.

Just thinking ahead also, but I've noticed the dependencies for the Avro serializer project uses `ME.Loging.Abstractions` with various versions depending on the framework version. The problem is that the `Confluent.SchemaRegistry` package transitively pulls in minimum version `8.x` of the `ME` packages, which will cause issues for the current setup.

So either the projects would all need to to target minimum `v8` of those packages, or there needs to be a new project that will only allow `net8.0` and higher. Perhaps `SlimMessageBus.Host.Serialization.AvroSchemaRegistry`?

See https://github.com/zarusz/SlimMessageBus/blob/master/src/SlimMessageBus.Host.Serialization.Avro/SlimMessageBus.Host.Serialization.Avro.csproj#L16

zarusz:
@robertcoltheart, good catch. I think it should be fine to bump the ME packages to target 8 version minimum, due to the transitive dependency in [Confluent.SchemaRegistry](https://www.nuget.org/packages/Confluent.SchemaRegistry?dependencies-body-tab).

image

Aside from that, I wonder if it is useful to have a base Avro serializer (current plugin), and have another plugin that layers schema registry functionality on top of it? I sense that users typically would use Avro with SchemaRegistry, so perhaps just evolving the existing plugin would make the most sense.

What do you think?

robertcoltheart:
Yes, typical business use case is to use a schema registry with Avro. You could still make it optional to use a schema registry with the changes I have in mind however.

For #372, is there any further changes you think are needed to pass the path to the serializer? It's best to put this change through first before I dive into the schema registry support.

zarusz:
For #372 I can merge this within a day. If you want to progress in the meantime feel free to start from that feature branch and then a simple rebase should suffice when I merge to master.

zarusz:
#372 has been merged to master, feel free to rebase when it's convenient.

zarusz:
hey @robertcoltheart are you still actively working on this?

robertcoltheart:
Sorry yes I've been quite busy and will get back to looking at the next part of this which would be to use the schema registry for Avro.

PR #376, opened 2025-03-04
[Host.Outbox] Clean up in batches

EtherZa:
Changes:
* Extracts `OutboxCleanUpTask` from `OutboxSendingTask`
* Clean up routing occurs outside of send loop and on SMB start up
* Clean up occurs in recurring batches until `Outbox` table has been cleaned
* Clean up occurs with a low `deadlock_priority` so as to kill service instead of publishing queries


zarusz:
Hi @EtherZa,

This is a valuable enhancement—great work!

Regarding the outbox integration tests, I wanted to share some context: our integration tests utilize a shared Azure SQL instance, meaning all test runs access the same database. Previously, when tests relied on the SQL Server container image provided by the build agent, we encountered significant unresponsiveness, leading to flaky outbox test results. I hope this insight is helpful. Please let me know if there's anything I can assist you with.​

EtherZa:
Hi @zarusz. That was quite the adventure but should hopefully be it. Please let me know your thoughts when you have a chance.

EtherZa:
FYI: I'm reviewing the indexes. I think it can be better.

EtherZa:
Thanks @zarusz. I'm done. Please review when you are able to.

src/SlimMessageBus.Host.Outbox/Services/OutboxCleanUpTask.cs
```diff
@@ -0,0 +1,185 @@
+﻿#nullable enable
+namespace SlimMessageBus.Host.Outbox.Services;
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
```
zarusz, line 6: These should come from implicit usings, so please remove them.

src/SlimMessageBus.Host.Outbox/Services/OutboxCleanUpTask.cs
```diff
@@ -0,0 +1,185 @@
...
+
+using Microsoft.Extensions.Hosting;
+
+using SlimMessageBus;
```
zarusz, line 10: This using seems redundant

src/SlimMessageBus.Host.Outbox/Services/OutboxCleanUpTask.cs
```diff
@@ -0,0 +1,185 @@
...
+
+public class OutboxCleanUpTask<TOutboxMessage, TOutboxMessageKey> : IMessageBusLifecycleInterceptor, IAsyncDisposable
+    where TOutboxMessage : OutboxMessage<TOutboxMessageKey>
+{
+    private readonly ILogger<OutboxCleanUpTask<TOutboxMessage, TOutboxMessageKey>> _logger;
+    private readonly OutboxSettings _outboxSettings;
+    private readonly TimeProvider _timeProvider;
+    private readonly IHostApplicationLifetime _hostApplicationLifetime;
+    private readonly IServiceProvider _serviceProvider;
+    private readonly SemaphoreSlim _slim;
```
zarusz, line 20: how about _semaphore?

src/SlimMessageBus.Host.Outbox/Services/OutboxCleanUpTask.cs
```diff
@@ -0,0 +1,185 @@
...
+    private async Task Startup()
+    {
+        await _slim.WaitAsync(_hostApplicationLifetime.ApplicationStopping);
+        try
+        {
+            Debug.Assert(_executingTask == null);
+            Debug.Assert(_cts == null);
+
+            _cts = CancellationTokenSource.CreateLinkedTokenSource(_hostApplicationLifetime.ApplicationStopping);
+            _executingTask = CleanUpLoop(_cts.Token);
```
zarusz, line 144: Wouldn't it be more optimal to use the [System.Threading.Timer](https://learn.microsoft.com/en-us/dotnet/standard/threading/timers) primitive to ask it to run with a specific interval, and do cleanup at those moments, and avoid a `Task.Delay()`?

src/Tests/SlimMessageBus.Host.Outbox.Test/SlimMessageBus.Host.Outbox.Test.csproj
```diff
@@ -2,7 +2,11 @@
 
 	<Import Project="../Host.Test.Properties.xml" />
 
-	<ItemGroup>
+  <ItemGroup>
```
zarusz, line 5: Ident, reformat

zarusz:
Nice work, @EtherZa! I’ve added a suggestion and a few minor comments - let me know what you think.

src/SlimMessageBus.Host.Outbox/Services/OutboxCleanUpTask.cs
EtherZa: Done.

src/SlimMessageBus.Host.Outbox/Services/OutboxCleanUpTask.cs
EtherZa: Done.

src/Tests/SlimMessageBus.Host.Outbox.Test/SlimMessageBus.Host.Outbox.Test.csproj
EtherZa: Done.

src/SlimMessageBus.Host.Outbox/Services/OutboxCleanUpTask.cs
EtherZa: No, they pretty much do the same thing. [`Task.Delay`](https://github.com/dotnet/runtime/blob/d17420560540e4184a78f431de8341f747068c4f/src/libraries/System.Private.CoreLib/src/System/Threading/Tasks/Task.cs#L5809) (at least when using [`TimeProvider`](https://github.com/dotnet/runtime/blob/d17420560540e4184a78f431de8341f747068c4f/src/libraries/Common/src/System/TimeProvider.cs#L11)) uses a [`TimerQueueTimer`](https://github.com/dotnet/runtime/blob/d17420560540e4184a78f431de8341f747068c4f/src/libraries/System.Private.CoreLib/src/System/Threading/Timer.cs#L451) under the hood.

image

which is the same implementation that [System.Threading.Timer](https://github.com/dotnet/runtime/blob/d17420560540e4184a78f431de8341f747068c4f/src/libraries/System.Private.CoreLib/src/System/Threading/Timer.cs#L908) uses

image

EtherZa:
> Nice work, @EtherZa! I’ve added a suggestion and a few minor comments - let me know what you think.

Thanks. I have made the changes you requested and added some detail arounnd `Task.Delay` with `TimeProvider`. tldr; it's the same as `System.Threading.Timer` :)

src/SlimMessageBus.Host.Outbox/Services/OutboxCleanUpTask.cs
zarusz: Ok then

EtherZa:
Hi @zarusz. Do you have any idea when this may escape to nuget.org?

EtherZa:
That sounds like a superb reason! Have a fantastic time

Issue #382, opened 2025-04-05
[Host.Outbox.Sql] Schema migration may result in dropped indexes being added again


EtherZa:
### Issue:
A race condition present when running multiple instancs of the application (with multiple schema migratiosn happening conccurently), may result in indexes being added after having been dropped.

Process | Event | Result 
--|--|--
1 | Retrieve status of `20230120000000_SMB_Init` | not yet deployed
2 | Retrieve status of `20230120000000_SMB_Init` | not yet deployed
1 | Deploy `20230120000000_SMB_Init` | add unused indexes
1 | Retrieve status of `20240503000000_SMB_Optimizations` | not yet deployed
1 | Deploy `20240503000000_SMB_Optimizations` | drop unused indexes
2 | Deploy `20230120000000_SMB_Init` | add unused indexes
2 | Retrieve status of `20240503000000_SMB_Optimizations` | deploy step
2 | Skip deploying `20240503000000_SMB_Optimizations` | 'dropped' indexes are retained

### Proposal:
Wrap schema migration in a distributed lock managed by SQL Server (sp_appgetlock).

zarusz:
Closed via #383

PR #383, opened 2025-04-05
#382 Wrap outbox schema migration in distributed lock

EtherZa:
Wraps a schema migration in a distributed app lock ([sp_appgetlock](https://learn.microsoft.com/en-us/sql/relational-databases/system-stored-procedures/sp-getapplock-transact-sql?view=sql-server-2016)) to ensure that a single process completes the migration. 

Resolves a race condition whereby indexes were added and removed by an instance only to be added back again by a second.

* No functional change is included.
* SlimMessageBus.Host.Outbox.Sql.Tests infrastructure scaffolding has been tweaked to reuse the SQL Server resource for a reduced test execution time.


zarusz:
Well done, thank you @EtherZa!

PR #389, opened 2025-04-12
Add Context accessor

t11omas:
Added a simple approach to setup IConsumerContextAccessor


zarusz:
@t11omas  what's the plan here?

Issue #392, opened 2025-04-15
[Host.Outbox.PostgreSql] Outbox provider for PostgreSql


EtherZa:
There was some chatter around 6 months ago in #311 around the posibility of a PostgreSql provider for the outbox. Unfortunately that seems to have died. To that end, please find the incoming PR.

zarusz:
Closed via #393 
Preview package pushed 3.2.0-rc.

PR #393, opened 2025-04-15
[Host.Outbox.PostgreSql] Outbox provider for PostgreSql

EtherZa:
A `Host.Outbox.PostgreSql[.DbContext]` implementation of the `Host.Outbox.Sql[.DbContext]` provider for use with PostgreSQL.

[Outbox message by reference](https://github.com/zarusz/SlimMessageBus/commit/54e6785e65a9d0154e57e82876e0c3d853dc2745)
* Reverted #325 `HasId` interface in favour of using object references to abstract the provider implementation from the `OutboxSendingTask`. This has been bundled as its own commit in the PR for traceability/review.

https://github.com/zarusz/SlimMessageBus/issues/392 [Host.Outbox.PostgreSql](https://github.com/zarusz/SlimMessageBus/commit/e6e3841060334dce8f8a851f339d088b1cb2c20e)
* Added `SlimMessageBus.Host.Outbox.PostgreSql` and `SlimMessageBus.Host.Outbox.PostgreSql.DbContext` as flavours of the `Sql Server` implementations
* Updated `Sample.OutboxWebApi` to select the target provider via enum selection
* Updated documentation to include references to both implementations
* Refactored `SqlOutboxTemplate` and its consumers to use an interface instead of a direct reference to the class

-sorry, this is a fairly chunky submission


zarusz:
@EtherZa As always — fantastic contribution and solid direction!

When you get a chance, could you take a quick look at the SonarCloud warnings? Some of them might be false positives, but a quick review could help fine-tune things.

I'll be doing a more in-depth review over the next few days.

Thanks again for your great work!

EtherZa:
I've gone through the SonarCloud complaints and resolved most of them while hiding the SQL injection concern. 

The outbox table/schema names are being injected into the migration scripts, which does pose a potential injection point, but as it would need to be manipulated by the same audience that uses it, I feel that the concern becomes moot (SQL Server migration has a similar implementation).

Code duplication is due to "copying" code from the SQL outbox implementation. I looked at abstracting this into a common library, but the SQL transport stub makes it a little scrappy. I chose to rather "reinvent the wheel" in the PostgreSql libs to keep the code together with minor tweaks where required.

src/SlimMessageBus.Host.Outbox.PostgreSql/PostgreSqlOutboxMessage.cs
```diff
@@ -0,0 +1,21 @@
+﻿namespace SlimMessageBus.Host.Outbox.PostgreSql;
+
+public class PostgreSqlOutboxMessage : OutboxMessage
+{
+    public Guid Id { get; set; }
+
+    public override string ToString()
+    {
+        return Id.ToString();
```
zarusz, line 9: Perhaps just =>

src/SlimMessageBus.Host.Outbox.Sql/SqlOutboxAdminMessage.cs
```diff
@@ -1,6 +1,14 @@
 ﻿namespace SlimMessageBus.Host.Outbox.Sql;
 
-public class SqlOutboxMessage : OutboxMessage<Guid>;
+public class SqlOutboxMessage : OutboxMessage
+{
+    public Guid Id { get; set; }
+
+    public override string ToString()
+    {
+        return Id.ToString();
```
zarusz, line 9: Perhaps =>

zarusz (approved):
@EtherZa looks good! 

Do you want update the top level README.md with the plugin reference (in the table) and  add a Postgre outbox call out in the README.md meant for NuGet packages? I could also add that here myself.

src/SlimMessageBus.Host.Outbox.PostgreSql/PostgreSqlOutboxMessage.cs
EtherZa: done

src/SlimMessageBus.Host.Outbox.Sql/SqlOutboxAdminMessage.cs
EtherZa: done

EtherZa:
Thanks @zarusz. I have made the two changes and updated the README.md. Please let me know if there is anything else you would like tweaked, otherwise it's good to go from my end.

PR #394, opened 2025-04-15
[Host] Ability to add assembly or type redirect for MessageType header

zarusz:
- Improved documentation for the `MessageType` header, including usage scenarios and the type redirection use case.  
- Introduced the `IAssemblyQualifiedNameMessageTypeResolverRedirect` interface to support type redirection with the default `AssemblyQualifiedNameMessageTypeResolver`.


zarusz:
@EtherZa could you please make a review?

EtherZa (approved):
Looks good :)

Issue #269, opened 2024-06-22
[Host.AmazonSQS] New transport for Amazon SQS and SNS


zarusz:
Please bring in [Amazon SQS](https://aws.amazon.com/sqs/) as another transport to SMB.

Proposed name of the NuGet and project:
`SlimMessageBus.Host.AmazonSQS`

https://docs.aws.amazon.com/sdk-for-net/v3/developer-guide/csharp_sqs_code_examples.html
https://docs.aws.amazon.com/sdk-for-net/v3/developer-guide/sqs-apis-intro.html

t11omas:
Hi,  Yes please create an empty plugin, and we will start to look at this.

zarusz:
Yup, will report when I have something for you.

zarusz:
I have started the SQS implementation. Once I have the first version working (it should be soon), then I will post here.

What do you think about having a combined transport for SQS and SNS in one transport?

t11omas:
I think that is a decent idea since they are quite closes related

zarusz:
Good, how about the name of the NuGet?

SlimMessageBus.Host.AmazonSQS is what I was thinking initially, when we combine SQS and SNS how about the package name?

Stick with: SlimMessageBus.Host.AmazonSQS?
Insicate its both: SlimMessageBus.Host.AmazonSQSSNS?

t11omas:
I don't have a strong preference, but I would say SlimMessageBus.Host.AmazonSQS looks better

zarusz:
For your information, I have a branch with the Amazon SQS provider that is mostly built. There are a couple of minor features (see ToDos) that I will work on when I have time:
https://github.com/zarusz/SlimMessageBus/tree/feature/amazon_sqs

Once I have all the needed features it will be made available as a pre-release in v3.

Feedback welcome.

zarusz:
The package is here
https://www.nuget.org/packages/SlimMessageBus.Host.AmazonSQS

t11omas:
Hi,
Sorry for the slow response, we have been using the SQS extension for consuming, and its working nicely.  Thanks so much for that.

Is there any update on when support for publishing to SNS will be introduced?

zarusz:
I need to find the time for SNS (unless the community is willing to take that on).

Also, it helps prioritize SNS when I know folks need it.

zarusz:
Just a heads-up — I've started working on adding Amazon SNS support. I'll share an update here once a preview version is ready.

[...]

PR #404, opened 2025-05-29
[Host.AmazonSQS] Add SNS support

zarusz:
As discussed in #269, I am adding the SNS transport to the `SlimMessageBus.Host.AmazonSQS` plugin.
These AWS services are typically used together, so it makes sense to keep this in one plugin.


src/SlimMessageBus.Host.AmazonSQS/ClientFactory/StaticCredentialsSnsClientProvider.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace SlimMessageBus.Host.AmazonSQS;
+
+public class StaticCredentialsSnsClientProvider(AmazonSimpleNotificationServiceConfig config, AWSCredentials credentials)
```
jaap-van-yperen, line 3: would be nice to have a client provider which uses the environment for credentials. im hosting my app on fargate, which provides credentials through env vars

i've added two of these to my project, one for sns and one for sqs. (even region can be an env var, but you get the drift)
`
    public AmbientCredentialsSnsClientProvider(IOptions<SlimMessageBusOptions> options)
    {
        _client = new AmazonSimpleNotificationServiceClient(Amazon.RegionEndpoint.GetBySystemName(options.Value.Region));
    }

`

zarusz: Good idea. I have incorporated the ambient credentials as the default option. The `UseRegion()` does not have to be provided (perhaps it will be pulled from the host ENV variables).

PR #421, opened 2025-07-26
[Host.Redis] prevent RedisListCheckerConsumer from crashing on disconnect 

### LINKED ISSUE #419: Redis consumer doesn't process events after reconnect
Hello

the issue appears when the consumer app briefly loses the connection to redis. After that no events are processed by the consumer until you restart it.

to reproduce:

- start producer and consumer
- send an event (works)
- close and open the redis connection 
- send an event - the consumer won't consume/process the event anymore until you restart it

Normally `RedisListCheckerConsumer` prints logs like this if it's working:

```
[11:01:52 VRB] Checking keys...
[11:01:52 VRB] Performing delay since no new items arrived
```

After the disconnect/reconnect `RedisListCheckerConsumer` doesn't log anything

producer:

        builder.Services.AddSlimMessageBus(mbb => {
            mbb.WithProviderRedis(settings =>
                settings.ConnectionString = "<connectionString>";
            mbb.AddServicesFromAssembly(Assembly.GetExecutingAssembly());

            mbb.Produce<TenantCreated>(producerBuilder => { producerBuilder.DefaultQueue("tenantCreated"); });

            mbb.AddJsonSerializer();
        });

consumer:

        builder.Services.AddSlimMessageBus(mbb => {
            mbb.WithProviderRedis(settings =>
                settings.ConnectionString = "<connectionString>";
            mbb.AddServicesFromAssembly(Assembly.GetExecutingAssembly());

            mbb.Consume<TenantCreated>(producerBuilder => { producerBuilder.Queue("tenantCreated");   });

            mbb.AddJsonSerializer();
        });

Version: 3.3.0

cyril265:
fixes #419 by catching exceptions
It retries the next queue "automatically" because it's inside the loop.
Draft because I didn't add a test case. Not really sure how to simulate a disconnect and reconnect inside a test.

### DIFF
--- a/src/SlimMessageBus.Host.Redis/Consumers/RedisListCheckerConsumer.cs
+++ b/src/SlimMessageBus.Host.Redis/Consumers/RedisListCheckerConsumer.cs
@@ -69,7 +69,16 @@ public class RedisListCheckerConsumer : AbstractConsumer, IRedisConsumer
             {
                 var queue = _queues[queueIndex];
 
-                var value = await _database.ListLeftPopAsync(queue.Name).ConfigureAwait(false);
+                var value = RedisValue.Null;
+                try
+                {
+                    value = await _database.ListLeftPopAsync(queue.Name).ConfigureAwait(false);
+                }
+                catch (Exception e) when (e is RedisConnectionException or RedisTimeoutException)
+                {
+                    Logger.LogWarning(e, "Redis connection error occurred while checking keys");
+                    await Task.Delay(1_000).ConfigureAwait(false);
+                }
                 if (value != RedisValue.Null)
                 {
                     Logger.LogDebug("Retrieved value on queue {Queue}", queue.Name);
--- a/src/SlimMessageBus.Host.Redis/Consumers/RedisListCheckerConsumer.cs
+++ b/src/SlimMessageBus.Host.Redis/Consumers/RedisListCheckerConsumer.cs
@@ -74,10 +74,13 @@ public class RedisListCheckerConsumer : AbstractConsumer, IRedisConsumer
                 {
                     value = await _database.ListLeftPopAsync(queue.Name).ConfigureAwait(false);
                 }
-                catch (Exception e) when (e is RedisConnectionException or RedisTimeoutException)
+                catch (Exception e)
                 {
-                    Logger.LogWarning(e, "Redis connection error occurred while checking keys");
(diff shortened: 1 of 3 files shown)


src/SlimMessageBus.Host.Redis/Consumers/RedisListCheckerConsumer.cs
```diff
@@ -69,7 +69,16 @@ protected async Task Run()
             {
                 var queue = _queues[queueIndex];
 
-                var value = await _database.ListLeftPopAsync(queue.Name).ConfigureAwait(false);
+                var value = RedisValue.Null;
+                try
+                {
+                    value = await _database.ListLeftPopAsync(queue.Name).ConfigureAwait(false);
+                }
+                catch (Exception e) when (e is RedisConnectionException or RedisTimeoutException)
```
zarusz, line 77: I would let it retry and delay endlessly upon any exception (not only redis exceptions).

src/SlimMessageBus.Host.Redis/Consumers/RedisListCheckerConsumer.cs
```diff
@@ -69,7 +69,16 @@ protected async Task Run()
...
-                var value = await _database.ListLeftPopAsync(queue.Name).ConfigureAwait(false);
+                var value = RedisValue.Null;
+                try
+                {
+                    value = await _database.ListLeftPopAsync(queue.Name).ConfigureAwait(false);
+                }
+                catch (Exception e) when (e is RedisConnectionException or RedisTimeoutException)
+                {
+                    Logger.LogWarning(e, "Redis connection error occurred while checking keys");
+                    await Task.Delay(1_000).ConfigureAwait(false);
```
zarusz, line 80: Maybe pass the cancallationToken too?

How about the delay TimeSpan setting is added to the `RedisMessageBusSettings`? Default being 1s.

zarusz: That could work, but then please update the property comment to mention that.

cyril265: I updated the property comment. Let me know if you prefer a separate property like you initially proposed. Something like  `QueueRetryDelay`

cyril265:
Hey @zarusz,
yep I also think it's ready now 😄

PR #447, opened 2025-12-02
[Host] Adds message header filtering for consumers

t11omas:
This pull request introduces a new feature for consumer message filtering based on message headers and, optionally, the raw transport message.

**Motivation and Goal:**
The primary goal is to provide a flexible and efficient way to route incoming messages to specific consumer invokers when multiple consumers subscribe to the same topic or queue. This allows for fine-grained message dispatching based on metadata in headers (e.g., `ResourceType`, `TenantId`, `MessageVersion`), avoiding the need to modify the underlying transport topology.

**New Features:**
- Introduces new `Filter` overloads in the consumer configuration builder (`AbstractConsumerBuilder`) to define predicates for message filtering.
- Supports filtering at both the consumer-level (`ConsumerSettings.Filter`) and the more specific per-invoker level (`IMessageTypeConsumerInvokerSettings.Filter`), with per-invoker filters taking precedence.
- Implements runtime evaluation of these filter predicates during message processing and invoker selection in `MessageProcessor`.
- Includes error handling for filter predicates: if a filter throws an exception, it is treated as a non-match, and a warning is logged.

**Noteworthy changes in usage for the end user:**
Users can now configure `.Filter()` on consumer registrations to conditionally execute consumers based on header values. This enables patterns like having multiple specialized consumers for the same message type on a single topic, each handling a subset of messages defined by header criteria.

**Documentation:**
- The `/docs/intro.md` documentation has been updated with a comprehensive section detailing "Consumer filtering," including its motivation, API usage, runtime behavior, precedence rules, examples, recommendations, and troubleshooting tips.

**Tests:**
- A new unit test, `When_Publish_Given_TwoConsumersOnSameTopic_WithDifferentFilters_Then_OnlyMatchingConsumerInvoked`, has been added to `MemoryMessageBusTests.cs` to validate the header-based consumer filtering functionality.


t11omas:
@zarusz I added the Signed off message on my commit but its still complaining, plus the build is failing, saying it can't find a file.  unfortunately I don't have time to investigate these :(

src/SlimMessageBus.Host.Configuration/Settings/ConsumerSettings.cs
```diff
@@ -50,4 +50,9 @@ private void CalculateResponseType()
     /// Enables the disposal of consumer instance after the message has been consumed.
     /// </summary>
     public bool IsDisposeConsumerEnabled { get; set; }
+
+    /// <summary> 
+    /// Optional predicate evaluated on arrival headers and transport message to decide if this invoker should be used. 
+    /// </summary> 
+    public Func<IReadOnlyDictionary<string, object>, object, bool> Filter { get; set; }
```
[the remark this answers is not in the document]
t11omas: Updated to use a delegate as per suggestion

src/SlimMessageBus.Host/Consumer/MessageProcessors/MessageProcessor.cs
```diff
@@ -185,35 +185,55 @@ protected Type GetMessageType(IReadOnlyDictionary<string, object> headers)
...
+    protected IEnumerable<IMessageTypeConsumerInvokerSettings> TryMatchConsumerInvoker(Type messageType, IReadOnlyDictionary<string, object> messageHeaders, object transportMessage)
     {
-        if (_singleInvoker != null)
-        {
-            // fallback to the first one
-            yield return _singleInvoker;
-        }
-        else
+
+        var invokers = _singleInvoker != null ? [_singleInvoker] : _invokers;
```
zarusz: We're introducing a memory alloc for the single element array here.
I understand its better to conbine the login into a one loop, but please consider the fact that most of the times users will have a single consumer declaration per type. More over the single invoker scenario also avoided foreach / IEnumerator creations and assignment check.

In other cases the consumer filtering might not be even used by users.

Even if we repeat that extra filter check code for that case, I think its better that way.

zarusz:
@t11omas I've finished my review, good job there!

Please check one more comment around memory alloc consideration.

Also, I still see two commits, so if you could squash it all into 1 commit - that keeps the history clean.

src/SlimMessageBus.Host/Consumer/MessageProcessors/MessageProcessor.cs
t11omas: Good point about the memory alloc.  I have reverted this to how it was before and added an ApplyFilter Method so it can be reused accordingly

zarusz:
@t11omas  thank you for your contribution. I will be releasing a stable build in some short time.
In the meantime there will be a rc build published.

PR #470, opened 2026-05-20
[Host.Kafka] Add KafkaLoopFailureInterceptor

tnmish:
### A new mechanism to intercept fatal exceptions occurring within the Kafka consumer group loop in KafkaGroupConsumer

A new interface IKafkaLoopFailureInterceptor was added. This interface defines a single method, OnFailureAsync, which is designed to be called when an unhandled exception terminates the main consumer group loop.
The ConsumerLoop method was modified. Inside its final catch (Exception e) block (which handles fatal errors), the code now creates a KafkaLoopFailureContext containing the consumer group name and the exception.
New Context Class (KafkaLoopFailureContext): A simple data class KafkaLoopFailureContext was introduced to encapsulate the information available when a fatal failure occurs: the Kafka consumer group ID and the exception itself.

Tests: Unit tests were added to verify that the IKafkaLoopFailureInterceptor.OnFailureAsync method is correctly invoked when a fatal exception occurs within the ConsumerLoop, ensuring the new functionality works as intended.

### DIFF
--- a/docs/provider_kafka.md
+++ b/docs/provider_kafka.md
@@ -17,6 +17,8 @@ Please read the [Introduction](intro.md) before reading this provider documentat
   - [Offset Commit](#offset-commit)
   - [Consumer Error Handling](#consumer-error-handling)
   - [Debugging](#debugging)
+- [Interceptors](#interceptors)
+  - [Consumer Loop Failure Interceptor](#consumer-loop-failure-interceptor)
 - [Deployment](#deployment)
 
 ## Underlying client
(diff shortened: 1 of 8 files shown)


zarusz:
@tnmish thanks for the contribution. This looks good. If you could also update the `provider_kafka.t.md` file and then run the  md post proces ps1 to refresh the `provider_kafka.md`.

tnmish:
> @tnmish thanks for the contribution. This looks good. If you could also update the `provider_kafka.t.md` file and then run the md post proces ps1 to refresh the `provider_kafka.md`.

done, commits squashed

zarusz:
@tnmish the integration tests were not picking up the secrets for connection. I have hopefully fixed that now. Can you rebase against master now?

tnmish:
First time rebasing onto upstream. Hope I got it right.

zarusz:
Thanks @tnmish for your contribution! 🎉

Issue #471, opened 2026-06-15
[Host.AmazonSQS]  Deleting By batch issue(s)


t11omas:
It is not 100% clear how our application got into this state, but there are 2 issues with Deleting of messages.

1. We ended up with an exception saying there was a duplicate messageId in the batch that was trying to delete.
2. Once the in memory list of messages to be deleted is greater than 10, it will just fail every time as there is an upper limit of 10 imposed from the AWS Api.

So I think the following needs to happen:

When adding message to the list to be deleted, only add unique message ids, and then when reading from the list, read in batches up to 10

https://github.com/zarusz/SlimMessageBus/blob/master/src/SlimMessageBus.Host.AmazonSQS/Consumer/SqsBaseConsumer.cs

zarusz:
Fixed by #473

PR #473, opened 2026-06-16
[Host.AmazonSQS] Enhance SQS message deletion robustness

t11omas:
Refactors SQS message deletion to use concurrent batches and a retry mechanism for transient failures. Messages from batches that fail to delete are now re-added to the internal processing queue for subsequent attempts, preventing re-processing of messages. A HashSet with a custom comparer ensures efficient tracking of messages to be deleted.

This is to address a bug, where the in memory list of messages can end up with duplicate message Ids, which results in an API exception, those messages are not removed from the in memory list, and every batch from this point on fails. The failure reason changes when the in memory list becomes more than 10, as AWS only allows batches of 10.

#471

### DIFF
--- a/src/SlimMessageBus.Host.AmazonSQS/Consumer/SqsBaseConsumer.cs
+++ b/src/SlimMessageBus.Host.AmazonSQS/Consumer/SqsBaseConsumer.cs
@@ -1,5 +1,8 @@
 ﻿namespace SlimMessageBus.Host.AmazonSQS;
 
+using System.Collections.Concurrent;
+
+
 abstract internal class SqsBaseConsumer : AbstractConsumer
 {
     private readonly ISqsClientProvider _clientProvider;
@@ -9,6 +12,7 @@ abstract internal class SqsBaseConsumer : AbstractConsumer
     private readonly int _visibilityTimeout;
     private readonly List<string> _messageAttributeNames;
     private readonly bool _isSubscribedToTopic;
+    private readonly SemaphoreSlim _semaphore = new SemaphoreSlim(10);
 
     private Task _task;
 
@@ -69,7 +73,7 @@ abstract internal class SqsBaseConsumer : AbstractConsumer
         return messageResponse.Messages;
     }
 
-    private async Task<bool> DeleteMessageBatchByUrl(string queueUrl, IReadOnlyCollection<Message> messages)
+    private async Task<List<string>> DeleteMessageBatchByUrl(string queueUrl, IReadOnlyCollection<Message> messages)
     {
         var deleteRequest = new DeleteMessageBatchRequest
         {
@@ -86,9 +90,16 @@ abstract internal class SqsBaseConsumer : AbstractConsumer
         }
 
         var deleteResponse = await _clientProvider.Client.DeleteMessageBatchAsync(deleteRequest, CancellationToken);
-
-        // ToDo: capture failed messages
-        return deleteResponse.Failed != null && deleteResponse.Failed.Count > 0;
+        var failedMessages = new List<string>();
+        if (deleteResponse.Failed != null)
+        {
+            foreach (var failed in deleteResponse.Failed)
+            {
+                failedMessages.Add(failed.Id);
+                Logger.LogWarning("Failed to delete message from Queue: {Queue}, SenderFault: {SenderFault}, Code: {Code}, Message: {Message}", Path, failed.SenderFault, failed.Code, failed.Message);
+            }
+        }
+        return failedMessages;
     }
 
     protected override Task OnStart()
@@ -109,8 +120,7 @@ abstract internal class SqsBaseConsumer : AbstractConsumer
     {
         var queueMeta = await MessageBus.TopologyCache.GetMetaWithPreloadOrException(Path, PathKind.Queue, CancellationToken);
         var queueUrl = queueMeta.Url;
-
-        var messagesToDelete = new List<Message>(_maxMessages);
+        var messagesToDelete = new Dictionary<string, Message>(_maxMessages);
 
         while (!CancellationToken.IsCancellationRequested)
         {
@@ -132,13 +142,13 @@ abstract internal class SqsBaseConsumer : AbstractConsumer
                             // ToDo: DLQ handling
                             break;
                         }
-                        messagesToDelete.Add(message);
+
+                        messagesToDelete[message.MessageId] = message;
                     }
                 }
                 if (messagesToDelete.Count > 0)
                 {
-                    await DeleteMessageBatchByUrl(queueUrl, messagesToDelete).ConfigureAwait(false);
-                    messagesToDelete.Clear();
+                    await DeleteMessages(messagesToDelete, queueUrl);
                 }
             }
             catch (TaskCanceledException)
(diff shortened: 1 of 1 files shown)


src/SlimMessageBus.Host.AmazonSQS/Consumer/SqsBaseConsumer.cs
```diff
@@ -153,6 +159,38 @@ protected async Task Run()
...
+        {
+            await semaphore.WaitAsync();
+
+            try
+            {
+                await DeleteMessageBatchByUrl(queueUrl, batch).ConfigureAwait(false);
+            }
+            catch
+            {
+                foreach (var item in batch)
```
zarusz: Add braces

src/SlimMessageBus.Host.AmazonSQS/Consumer/SqsBaseConsumer.cs
```diff
@@ -153,6 +159,38 @@ protected async Task Run()
...
+            {
+                semaphore.Release();
+            }
+        });
+
+        await Task.WhenAll(tasks);
+
+        messagesToDelete.Clear();
+
+        foreach (var item in failed)
```
zarusz: Braces

src/SlimMessageBus.Host.AmazonSQS/Consumer/SqsBaseConsumer.cs
```diff
@@ -153,6 +159,38 @@ protected async Task Run()
...
     }
 
+    private async Task DeleteMessages(HashSet<Message> messagesToDelete, string queueUrl)
+    {
+        var failed = new ConcurrentBag<Message>();
+        var semaphore = new SemaphoreSlim(10);
+
+        var tasks = messagesToDelete.Chunk(10).Select(async batch =>
+        {
+            await semaphore.WaitAsync();
```
zarusz: I wonder if we need the semaphore here.

Is the idea to delete the chunks in pararell but up to 10 at a time?
Can't we simply process all the chunks in pararell without a sempore?

t11omas:
Yes to delete in parallel up to 10 at a time.


I did use the parallel extension methods but they were not available in .net standard, which the project targets.

So I went for a single approach rather that different ones based on the .net version.

The reason I added a semaphore was to avoid bombarding the AWS api should the messagesToDelete list be large

Happy todo whatever you feel is best though

Sent from Outlook for iOS<https://aka.ms/o0ukef>
________________________________

zarusz:
@t11omas  could you also consider these?


The two important issues are both in SqsBaseConsumer.cs:
Partial SQS batch-delete failures are ignored.
DeleteMessageBatchAsync can return HTTP 200 while individual entries fail. The PR’s DeleteMessageBatchByUrl still only returns true/false, and DeleteMessages ignores that return value, clears messagesToDelete, and loses the failed entries. That means processed messages whose delete failed can be redelivered later.

Retrying by MessageId is unsafe.
The new HashSet<Message> compares only MessageId, but SQS deletion depends on the latest ReceiptHandle. If a delete throws, then the message becomes visible and is received again, it will have the same MessageId but a newer receipt handle. The hash set keeps the old object/old handle, so later delete retries may use a stale receipt handle and fail to actually delete the message.

Worth improving:
Track deletion retry entries by batch entry id / receipt handle, not just message id.
Have DeleteMessageBatchByUrl return the actual failed entry ids, then re-add only those messages to the retry set.
Log failed batch entries with Code, Message, and SenderFault; sender faults probably should not be retried forever.
Remove nullable annotations in Message? or enable nullable context, because the PR adds CS8632 warnings.

Semaphore:
Could it be re-used to avoid allocs with every message delete/completion round?
It might be more desired to limit to up to 10 concurent requests per consumer, not only to 10 per message delete batch.

src/SlimMessageBus.Host.AmazonSQS/Consumer/SqsBaseConsumer.cs
t11omas: added

src/SlimMessageBus.Host.AmazonSQS/Consumer/SqsBaseConsumer.cs
t11omas: added

t11omas:
> @t11omas could you also consider these?
> 
> The two important issues are both in SqsBaseConsumer.cs: Partial SQS batch-delete failures are ignored. DeleteMessageBatchAsync can return HTTP 200 while individual entries fail. The PR’s DeleteMessageBatchByUrl still only returns true/false, and DeleteMessages ignores that return value, clears messagesToDelete, and loses the failed entries. That means processed messages whose delete failed can be redelivered later.
> 
> Retrying by MessageId is unsafe. The new HashSet compares only MessageId, but SQS deletion depends on the latest ReceiptHandle. If a delete throws, then the message becomes visible and is received again, it will have the same MessageId but a newer receipt handle. The hash set keeps the old object/old handle, so later delete retries may use a stale receipt handle and fail to actually delete the message.
> 
> Worth improving: Track deletion retry entries by batch entry id / receipt handle, not just message id. Have DeleteMessageBatchByUrl return the actual failed entry ids, then re-add only those messages to the retry set. Log failed batch entries with Code, Message, and SenderFault; sender faults probably should not be retried forever. Remove nullable annotations in Message? or enable nullable context, because the PR adds CS8632 warnings.
> 
> Semaphore: Could it be re-used to avoid allocs with every message delete/completion round? It might be more desired to limit to up to 10 concurent requests per consumer, not only to 10 per message delete batch.

Updated the PR based on your comments.

However the one thing I didn't address or agree with in the first instance was limit the concurrent requests per consumer to match the delete batch.  The deleting by batch is a framework implementation details, that IMO shouldn't be imposed limits on consumers

zarusz:
Hey @t11omas  looks much better - thanks. 

There is still this ReceiptHandle issue, that we'd likely need to track the message to be deleted. Perhap your original issue was due to the message being redelivered (same MessageID but different handle).

Here is what my coding agent longer elaboration on the matter 👇 

Let me know if you can accomodate, in this PR, otherwise I can do this in a separate PR myself.
In any case please be sure to squash the PR into one commit before I can merge.

----

Yes. The latest commit addresses the big partial-failure ask: `DeleteMessageBatchAsync` failures are now read from `deleteResponse.Failed`, logged, and retained for retry.

I still see one correctness concern worth asking them to fix:

`SqsBaseConsumer.cs` still keys pending deletes by `MessageId`:

```csharp
messagesToDelete[message.MessageId] = message;
```

and uses `MessageId` as the SQS batch entry id. For SQS, deletion is tied to the receipt handle, and receipt handles are per receive. Standard queues can redeliver the same logical message, so the same `MessageId` may appear with a different `ReceiptHandle`. With the current dictionary, those deliveries collapse into one entry. I’d prefer keying pending deletes by `ReceiptHandle`, and using a generated unique batch entry id that maps back to the message/receipt handle for failed-entry handling.


----

Yes, it can still pose an issue, because SQS does not delete “the message by `MessageId`.” It deletes by `ReceiptHandle`.

In this PR, `MessageId` is only used as the batch entry `Id` and as the dictionary key. The actual delete request still sends:

```csharp
ReceiptHandle = message.ReceiptHandle
```

So the problem is not that AWS receives `MessageId` instead of `ReceiptHandle`; the problem is that the in-memory retry tracking can collapse multiple receives of the same logical SQS message into one pending delete entry.

If the consumer processes the same SQS message again after a previous delete failed, it may get:

```text
MessageId:      abc
ReceiptHandle:  old-handle
```

then later:

```text
MessageId:      abc
ReceiptHandle:  new-handle
```

Using `messagesToDelete[message.MessageId] = message` means only one of those survives. If it keeps the newest one, that is usually better than the original `HashSet` version. But it still loses the fact that two processing attempts completed and two delete attempts may have distinct outcomes.

Semantically, if your model is “once we processed this logical message successfully, delete whichever latest receipt handle we have,” then keying by `MessageId` is defensible. It is probably okay for the common single-consumer/sequential loop case, because a later receive overwrites the older handle with the newer one.

The caveat is SQS edge cases:

- `MessageId` is not the delete token; `ReceiptHandle` is.
- Batch entry `Id` only needs to be unique within the batch, so using `MessageId` is okay only if the batch cannot contain duplicates.
- If duplicates with the same `MessageId` can exist in the pending dictionary, the current code cannot represent them independently.
- For FIFO queues or unusual redelivery/concurrency cases, collapsing by `MessageId` may hide one completed delivery attempt’s delete result.

So I’d soften the review concern from “definite bug” to “robustness issue.” If this consumer processes receives sequentially and only has one polling loop per queue instance, the current PR is likely acceptable. The more correct SQS-shaped implementation would track by `ReceiptHandle` and use a separate batch entry id, but I would not block the PR solely on this unless you want maximum correctness around redelivery edge cases.

----

t11omas:
> ```cs
> messagesToDelete[message.MessageId] = message;
> ```

We don't want duplicate messageIds, and the ReceiptHandle can become stale, so if the same message id comes into to be processed, we want to have the latest ReceiptHandle, we wont want to hold on to state data.  So the ReceiptHandle is being Upserted.  I don't think I follow what you are suggesting is a problem here.

zarusz:
Ok got it @t11omas. We will preserve the last message by ID which should have the latest handle.
Happy to marge the PR. Thanks for you contribution!

PR #475, opened 2026-06-17
Fix outbox failed batch retry loop

### LINKED ISSUE #459: [Host.RabbitMq] Outbox retries exhaust quickly when broker is down
As pointed out in #458:

Outbox retries exhaust quickly when broker is down. When the RabbitMQ broker is unreachable, EnsureChannel() throws immediately on every dispatch attempt. The outbox dispatcher retries in rapid succession, consuming all MaxDeliveryAttempts within seconds rather than backing off. A short-circuit mechanism in the outbox dispatcher — skipping dispatch attempts while the transport reports unhealthy and resuming when the connection recovers — would prevent this. This is a transport-agnostic outbox concern and should be addressed separately. I believe I saw a branch that adds the short-circuit but just wanted to call that out here as I noticed this issue in local testing.

zarusz:
## Summary
- Fixes #459 by stopping failed outbox batches from immediately re-entering the same sending pass.
- Failed messages still have delivery attempts recorded, but the next retry is governed by the outbox polling/lock timing instead of a tight same-instance loop.
- Adds core outbox unit coverage and PostgreSQL/Testcontainers integration coverage for the retry behavior.

## Tests
- dotnet test src\Tests\SlimMessageBus.Host.Outbox.Test\SlimMessageBus.Host.Outbox.Test.csproj --no-build -v minimal -m:1
- dotnet test src\Tests\SlimMessageBus.Host.Outbox.PostgreSql.Test\SlimMessageBus.Host.Outbox.PostgreSql.Test.csproj --no-build -v minimal -m:1

### DIFF
--- a/src/SlimMessageBus.Host.Outbox/Services/OutboxSendingTask.cs
+++ b/src/SlimMessageBus.Host.Outbox/Services/OutboxSendingTask.cs
@@ -237,6 +237,7 @@ internal class OutboxSendingTask<TOutboxMessage>(
         const int defaultBatchSize = 50;
 
         var runAgain = outboxMessages.Count == _outboxSettings.PollBatchSize;
+        var failed = false;
         var count = 0;
 
         var aborted = new List<TOutboxMessage>(_outboxSettings.PollBatchSize);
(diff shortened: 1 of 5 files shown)


zarusz:
@cord-agencyroot do you want to review this PR? It should fix one of the issues you've reported previously. Thanks.

cord-agencyroot:
@zarusz looks good! thanks for fixing this

